PR #4, opened 2015-01-21
Make a new CAPI client for each update

richard-still-ft:
So that items cached in the CAPI are not used


matthew-andrews:
lgtm :+1:

PR #11, opened 2015-02-16
Enable blogposts on notifications feed

adgad:
/cc @richard-still-ft


adgad:
Do we need to backpopulate the missing ones somehow?

richard-still-ft:
You can do a week of updares via the rest call but that is an arbitrary restriction. I will tweak it to allow a longer time period

PR #34, opened 2015-05-11
Updated the realm for clarity. Added a new build icon.

richard-still-ft:


richard-still-ft:
It was the readme I updated not the realm. Stupid typo correction.

PR #36, opened 2015-05-22
Add basic functionality to patch a document with arbitrary annotations

i-like-robots:
TODO: support appending to the array of custom annotations (req. script)

Example usage:

```
curl -X PATCH 'http://localhost:5000/api/item?apiKey=123' -d '{ "id": "804c688b-ff1e-32bf-b6b6-ba0f6bb5a161", "index": "v1_api_v2", "type": "item", "nextAnnotation": "Hello World" }' -H 'Content-Type: application/json'
```


richard-still-ft:
:+1:

PR #79, opened 2015-07-02
Next 'thing' endpoint

matthew-andrews:
@wheresrhys


matthew-andrews:
not yet aware of mapping of tags…… just purely fails when not available (so don't use for deleting tags…… only use for updating of labels for now……)

PR #80, opened 2015-07-03
Refactor metadata so that only delete tags when we have to, normally just map

matthew-andrews:
This will help me implement mapping in the thing endpoint

cc @wheresrhys


wheresrhys:
What a massive can of worms all this retagging is

matthew-andrews:
ha yes. it's fun.

matthew-andrews:
hopefully this won't break stuff :see_no_evil:

PR #81, opened 2015-07-03
Expose mapped IDs

matthew-andrews:


matthew-andrews:
@wheresrhys it's done :-)

PR #97, opened 2015-08-06
Map London person to region

matthew-andrews:


tests/metadata-transformer-spec.js
```diff
@@ -216,4 +233,24 @@ describe('metadata transformer', function() {
...
 		});
 	});
+
+	it('should map London (the person) to London (the region)', function() {
+		var metadata = {
+			primaryTheme: londonPersonTag,
+			tags: [ londonPersonTag ],
+			people: [ londonPersonTag ]
+		};
+		var article = articleModel({
```
matthew-andrews: the metadata transformer, confusingly only does _some_ of the metadata transforms :see_no_evil: 

to get the whole lot need to pass through `articleModel`

PR #102, opened 2015-08-11
Make 7 more people special

keirog:
This’ll make them show in the branding area in stream cards (above the
time)


matthew-andrews:
:+1:

PR #103, opened 2015-08-12
Ads update

robin-marr-ft:
Add DFP sites and Zones for the majority of the nav


robin-marr-ft:
Hi @matthew-andrews

So I've broken mapping up by target taxonomy topic and use loadash merge to create the final object in map.js.

Couple of tests failed because dfp  site has changed to remove next and capitals so fixed those too.

Cheers
Rob

matthew-andrews:
lovely work, thanks! :+1:

PR #105, opened 2015-08-12
Use NBT for env vars

i-like-robots:
Use NBT to set env vars where possible and switch to more standard Harrier access and FT API keys.

Requires http://git.svc.ft.com/projects/NEXTPRIVATE/repos/config-vars/pull-requests/149/overview


richard-still-ft:
:+1:

PR #106, opened 2015-08-14
Configure the box before deploying

i-like-robots:
@richard-still-ft


i-like-robots:
screen shot 2015-08-14 at 11 18 39

\0/

PR #107, opened 2015-08-20
Except for a whitelist of OK tags, remove photo diary articles from all streams & myft feeds

matthew-andrews:


tests/metadata-transformer-spec.js
```diff
@@ -256,4 +306,28 @@ describe('metadata transformer', function() {
 			people: []
 		});
 	});
+
+	it.only('should do nasty things to the photo diary', function() {
```
i-like-robots: Remove the only before merge

PR #109, opened 2015-08-26
fix ampersands in capi v1

matthew-andrews:
@adambraimbridge


adambraimbridge:
:+1:  (after linting fix)

PR #110, opened 2015-08-26
podcasts

i-like-robots:
- Fetches multiple RSS feeds for each show in series
- For each feed it will take each episode, map it to the (V1) article model and insert them into ES in throttled batches
- It maps each episode to meta data and old Podhoster IDs to new Acast GUIDs
- It will attempt to update each episode up to 3 times
- It can reindex all episodes or only add new shows

screen shot 2015-09-03 at 12 50 05

Depends on https://github.com/Financial-Times/next-feature-flags-api/pull/326 and https://github.com/Financial-Times/next-ft-api-client/pull/48


server/podcast-feeds.js
```diff
@@ -0,0 +1,26 @@
...
+var args = require('minimist')(process.argv.slice(2));
+var PodcastUpdater = require('./processes/podcast-updater');
+
+var feeds = [
+	'http://rss.acast.com/ft-uk-general-election-2015'
+];
+
+var tasks = feeds.map(function(feed) {
+	return new PodcastUpdater({
+		elasticSearchURL: 'https://b1befb2bddf0fd14d4a5285ccd141646.eu-west-1.aws.found.io:9243/',
```
i-like-robots: This is my own cluster for testing

matthew-andrews:
One small question, have you checked what happens when the continuous updater encounters a podcast?

i-like-robots:
No, added to to-do

server/processes/podcast-updater.js
```diff
@@ -0,0 +1,195 @@
...
+			if (data.hits.total) {
+				return new Date(data.hits.hits[0]._source._lastUpdatedDateTime);
+			}
+		});
+};
+
+PodcastUpdater.prototype.fetchRssFeed = function() {
+	var parser = new FeedMe(true);
+
+	return fetch(this.options.rssFeedUrl, { timeout: 3000 })
```
matthew-andrews: error handling?

server/processes/podcast-updater.js
```diff
@@ -0,0 +1,195 @@
...
+		.then(function(response) {
+			return response.text();
+		})
+		.then(function(text) {
+			parser.on('error', function(err) {
+				throw err;
```
matthew-andrews: oh i guess that'll do

server/processes/podcast-updater.js
```diff
@@ -0,0 +1,195 @@
...
+				throw new Error(data.error);
+			}
+
+			console.log('' + id + ' updated to version ' + data._version + '.');
+			return Promise.resolve();
+		});
+};
+
+PodcastUpdater.prototype.error = function(error) {
+	console.error(error.message || error);
```
matthew-andrews: ^^ `raven.captureError` ?

matthew-andrews:
Looks good…  it would be good to think about how we monitor these to make sure they're working.  E.g. the health check for CAPI content will fail if it can't continuously update… a good test of quite a lot of the moving parts as we learned this week……

[Not necessary for this stage, but probably ought to be done before the migration from FT to Next is done]

server/processes/podcast-updater.js
i-like-robots: Pardon?

matthew-andrews: can we send these these errors to sentry :innocent:

matthew-andrews: or rather, should we?

i-like-robots:
Current plan is to run it manually a few times to verify the data is OK. Health checks and setting up the scheduler could be next?

server/processes/podcast-updater.js
i-like-robots: I think we should, maybe as mentioned below I can add monitoring etc. when this task requires scheduling?

matthew-andrews:
sounds good

PR #115, opened 2015-09-02
load data set once

commuterjoy:


commuterjoy:
```
var p = new Promise(function(resolve, reject) { 
    console.log('expensive thing');
    return [1, 2, 3] 
})

p.then(function(v) { console.log(v) });
p.then(function(v) { console.log(v) });
...

// or

var f = function (v) { return p.then(function (vv) { return vv.indexOf(v); }); }
f(1).then(function(n) { console.log(n); }) // 0
f(9).then(function(n) { console.log(n); }) // -1
```

commuterjoy:
Might have misinterpreted your question, but a Promise is repeatedly _thenable_ once it's resolved.

PR #116, opened 2015-09-03
/_search -> /item/_search

wheresrhys:
Unexpectedly there's no need to alter the query objects. I've verified all the routes still work (the tests don't seem to cover the bits of code concerned. Unit tests couldn't really - needs integration tests) @matthew-andrews


matthew-andrews:
:+1:

PR #123, opened 2015-09-21
Switch to use source URL as key for podcast metadata

i-like-robots:
Because the titles have and will change. Using the source URL should be nearly as readable.


andygnewman:
:+1:

PR #124, opened 2015-09-21
override for now

matthew-andrews:


i-like-robots:
Is the issue that this app is named 'Harrier' in the service registry but 'Harrier EU' everywhere else?

i-like-robots:
:ship:

matthew-andrews:
hmm wait

matthew-andrews:
ok now is ok

i-like-robots:
:shipit:

PR #131, opened 2015-10-01
Ensure otherTitles is an empty array

i-like-robots:
Podcasts are currently indexed with the `otherTitles` property as `[ {} ]` despite this being specified in the mapping as a string and not previously included in our model. This stops ES being an idiot.

@richard-still-ft


richard-still-ft:
+1

PR #137, opened 2015-10-06
Add schema for article v3 index

i-like-robots:
- Based on CAPI v2 with addition of titles, summaries, provenance and attachment fields.
- Attachments and provenance fields are for Next usage and not related to CAPI in any way.
- Annotations will be V1 for now.


matthew-andrews:
:+1:

PR #138, opened 2015-10-07
Workaround a data bug in Concordance API that break the P&G and Disney streams

matthew-andrews:


matthew-andrews:
I think they're going to try to fix this upstream…

matthew-andrews:
stale

PR #139, opened 2015-10-07
Map facebook (Topic) to Facebook Inc (org)

adgad:
/cc @matthew-andrews


matthew-andrews:
what could possibly go wrong.  :+1:

PR #141, opened 2015-10-08
v3 models

i-like-robots:
Adds V3 article model which:

- fetches item from CAPI V1 and V2
- assembles the article from those two sources
- if only a V1 response it will strip the content markup leaving only paragraphs
- if only a V2 response it will attempt to create V1 compatible metadata
- either way it will attempt to concord the metadata and store both V1 and V2 tag IDs


server/models/article-v3.js
```diff
@@ -0,0 +1,162 @@
...
+	return {
+		id: articleV1.item.id,
+		title: articleV1.item.title.title,
+		titles: [].concat(
+			articleV1.item.packaging.spHeadline,
+			// This can sometimes be an empty object because of an ES bug
+			// but don't worry, it will get filtered out later!
+			articleV1.item.editorial.otherTitles
+		),
+		byline: articleV1.item.editorial.byline,
```
matthew-andrews: i think if the content isn't in CAPI v2 then we shouldn't bother with trying to recover its byline from v1

server/models/article-v3.js
```diff
@@ -0,0 +1,162 @@
...
+		// Strip all HTML except paragraphs
+		bodyXML: articleV1.item.body.body.replace(/<(?!\s*\/?\s*p\b)[^>]*>/gi,''),
+		summaries: [
+			articleV1.item.summary.excerpt,
+			articleV1.item.editorial.subheading
+		],
+		publishedDate: articleV1.item.lifecycle.lastPublishDateTime,
+		mainImage: (() => {
```
matthew-andrews: same for main image as per byline

server/models/article-v3.js
```diff
@@ -0,0 +1,162 @@
...
+
+function normalizeItemV2(articleV2) {
+	return {
+		id: articleV2.id,
+		title: articleV2.title,
+		byline: articleV2.byline,
+		bodyXML: articleV2.bodyXML,
+		publishedDate: articleV2.publishedDate,
+		mainImage: { url: '', description: '' },
+		webUrl: articleV2.webUrl,
```
matthew-andrews: we actually probably will *prefer* the capi v1 `webUrl` over capi v2 becasue it has the content level in it

server/models/article-v3.js
```diff
@@ -0,0 +1,162 @@
...
+		.sort(function(a, b) {
+			// Descending order, longest to shortest
+			return b.length - a.length;
+		});
+}
+
+function applyModel(articleV1, articleV2) {
+	// Rather than write lots of ternary statements let's just make
+	// both APIs speak the same language.
+	let normalizedV1 = articleV1 ? normalizeItemV1(articleV1) : {};
```
matthew-andrews: suggest sticking a try catch around the v1 normalisation…  random bits of the article model go randomly missing sometimes

server/models/article-v3.js
matthew-andrews: probably simplest to deal with this exception separately

server/models/article-v3.js
```diff
@@ -0,0 +1,162 @@
...
+
+	if (article.summaries) {
+		article.summaries = cleanAndSortStrings(article.summaries);
+	}
+
+	// We'll continue to use V1 metadata if it's available...
+	let metadata = articleV1 ? metadataV1(articleV1) : metadataV2(articleV2);
+
+	// ... but, we'll try to store both V1 and V2 tag IDs for future.
+	metadata.then(function(concorded) {
```
matthew-andrews: i think you mean `metadata = metadata.then(…`

server/models/article-v3.js
i-like-robots: content level?

matthew-andrews: Spot the difference:-
screen shot 2015-10-08 at 15 41 11

screen shot 2015-10-08 at 15 41 04

server/models/article-v3.js
i-like-robots: I'll import a few thousand articles and see what breaks first

matthew-andrews: good idea

PR #145, opened 2015-10-12
Preresolve images

i-like-robots:
- Finds `<ft-content>` image set tags in the given XML
- Resolves each image set and then its first media resource
- Replaces the `<ft-content>` tags with image elements or removes any which could not be resolved


matthew-andrews:
looks nice :)

PR #151, opened 2015-10-13
Prefix spHeadline with kicker if there is one

matthew-andrews:
e.g. http://api.ft.com/content/items/v1/d644ddd6-718c-11e5-9b9e-690fdae72044


matthew-andrews:
It'd be nice if nice if it were easy to create tests for this kind of thing… i.e. not need to nock api responses :-/

PR #166, opened 2015-10-14
TMEs

matthew-andrews:
@wheresrhys — 

The new cluster is FAST!


server/controllers/tmes.js
```diff
@@ -0,0 +1,58 @@
...
+			// It really does timeout for organisations if this limit isn't high
+			timeout: 50000,
+			method: 'POST',
+			body: JSON.stringify({
+				size: 0,
+				aggs: {
+					'primary-sections': {
+						terms: {
+							field: `item.metadata.${req.params.taxonomy}.term.name`,
+							size: 40000
```
wheresrhys: Given it's so fast, may as well make this number bigger. I think it's fairly close to the actual number of organisations

matthew-andrews: false alarm was looking at the wrong place

PR #168, opened 2015-10-19
Stop podcasts

i-like-robots:
Acast unfortunately provided us with incorrect IDs for episodes migrated from the existing provider. We built our articles app integration based on the format of these IDs. 

We have now been provided with the correct IDs. If we use these then our articles app integration will break.


matthew-andrews:
:+1: gonna need to do some service resgitry stuff too, no?

matthew-andrews:
also need to be pulled into `es-interface` branch too

matthew-andrews:
in fact isn't it easier to **just** do this in the service registry?

i-like-robots:
Would updating the service registry stop the processes being started?

Our linting goalposts appear to have been moved too =S

matthew-andrews:
@i-like-robots — yep on subsequents deploys and the can stop the existing process manually

i-like-robots:
This was pointless.

PR #169, opened 2015-10-20
Primary tags in v3

i-like-robots:
... And a little cleanup.

- Removed promises from metadata mappings as they're no longer required.
- Store V2 annotation `directType` property instead of attempting to map it to a V1 equiv. taxonomy since we're storing the two datasets separately now.
- Add a boolean `primary` property to each metadata entry to indicate if it was defined as a `primarySection` or `primaryTheme` in V1 API.


matthew-andrews:
:+1:

PR #170, opened 2015-10-20
Podcasts V3

i-like-robots:
- Adds podcast model for V3 index
- Adds support for V3 index in updater process


matthew-andrews:
:+1:

PR #177, opened 2015-10-23
Missing stuff for v3

i-like-robots:
- Adds standFirst to summaries (they're not always the same as subheadings after all)
- Adds story packages from V1 to fix https://github.com/Financial-Times/next-harrier/issues/175


matthew-andrews:
:+1:

PR #188, opened 2015-10-28
If a tag has been deleted (e.g. Front page) it should be deleted from…

matthew-andrews:
… ‘primaryX’ as well … replacing with the genre, if an article has one


wheresrhys:
looks good

PR #192, opened 2015-11-04
[WIP] Author / brand transformation

andygnewman:
- [x] add Jonathan Margolis to authors who are also brands.
- [x] reorder author metadata object into alphabetical order.
- [ ] add attribute `hasHeadshot' to authors with headshots we want to use (in V3 transform only using external component - https://github.com/Financial-Times/n-headshot-mapping/pull/1)

TO DO 
- [ ] stop hack that overrides primaryTheme with author if author = brand (need to consult other streams (esp home page as would remove branding from cards)


andygnewman:
No longer relevant

PR #198, opened 2015-11-10
Point to the Elastic Search DNS so it falls over correctly

adgad:
/cc @i-like-robots @richard-still-ft @phamann


matthew-andrews:
yeah dont do this

Issue #196, opened 2015-11-10
Move transformations from article app into ingest


i-like-robots:
Transforms within the article app that are unopinionated about presentation should be performed as content is ingested to prevent repeated work.

- [x] https://github.com/Financial-Times/next-article/blob/master/server/transforms/relative-links.js
- [x] https://github.com/Financial-Times/next-article/blob/master/server/transforms/replace-ellipses.js
- [x] https://github.com/Financial-Times/next-article/blob/master/server/transforms/replace-hrs.js

We could also consider:

- [x] Removing newlines I.E. `/\r?\n|\r/`
- [x] Removing redundant elements such as empty paragraphs or line breaks between paragraphs

wheresrhys:
Following a meeting with the data team it'd be great if these transforms could also output statistics on the type of content contained in an article in something like a `contentStats` property.

PR #199, opened 2015-11-10
Remove newlines, empty paragraphs and redundant break tags 

i-like-robots:
The content we ingest from CAPI v2 can be filled with useless stuff we should not store or fix. This covers a few of those things.

https://github.com/Financial-Times/next-es-interface/issues/196


server/models/article-v3.js
```diff
@@ -103,6 +104,18 @@ function processArticleV1Body(articleV1) {
...
 }
 
+function processArticleV2Body(articleV2) {
+	return articleV2.bodyXML
+		// Remove all new lines
+		.replace(/\r?\n|\r/g, '')
+		// Remove empty paragraphs
+		.replace(/<p>\s*<\/p>/g, '')
+		// Remove useless break elements
+		.replace(/<\/p>\s*<br\/>\s*<p>/g, '</p></p>')
```
andygnewman: Have seen some useless & empty <strong> and <li> tags too that you could include here also.

matthew-andrews:
:+1:

PR #206, opened 2015-11-20
Also ban ‘NV’ — https://en.wikipedia.org/wiki/Naamloze_vennootschap

matthew-andrews:
@musthemag :+1: / :-1: ?


matthew-andrews:
@andygnewman (also adds `\b`'s to the beginning of all the checks so that it only trims words, not orgs who's names might end in these patterns — unlikely but could be embarrassing)

PR #211, opened 2015-11-23
Temporarily exclude podcasts from cts updater checks

matthew-andrews:


matthew-andrews:
@i-like-robots Any reason why V3 model code (and all associated stuff — continuous updater, the ‘help’ endpoints — https://ft-next-es-interface-eu.herokuapp.com/help) can't currently cope with Podcasts  …?

i-like-robots:
They don't share a much if any commonality in the way they fetch or transform the data. I have thought about going OO and having a base updater class we extend for both CAPI and Podcasts but that's probably over engineering.

PR #212, opened 2015-11-23
Exclude podcasts from continuous updater (**temporarily**)

matthew-andrews:


server/continuous-updater.js
```diff
@@ -18,32 +19,28 @@ require('isomorphic-fetch');
...
-			body: JSON.stringify({
-					fields: ['_lastUpdatedDateTime'],
-					sort: {
-						_lastUpdatedDateTime: 'asc'
-					},
-					query: {
-						bool: {
-							must_not: [
-								{
-									term: { 'item.metadata.sections.term.name': 'Podcasts' }
```
matthew-andrews: this only worked for v2, for v3 this meant i suspect the continuous updater was grinding to a halt……

matthew-andrews:
doesn't work

PR #214, opened 2015-11-24
Prefer webUrl from CAPI v1

matthew-andrews:


i-like-robots:
Probably worth a comment too :+1:

matthew-andrews:
good call

PR #228, opened 2015-12-14
Less v2

i-like-robots:


Procfile
```diff
@@ -1,10 +1,7 @@
 web: forever server/app.js
 v1Api: forever server/update-search.js --index v1_api_v2 --type item
-v2Api: forever server/update-search.js --index v2_api_v2 --type item --apiVersion 2
```
matthew-andrews: we still need this one i think because for notifications feed this triggers V3 ingests :)

README.md
```diff
@@ -23,7 +23,6 @@ Currently only supports V1 content.
 `update-search.js` keeps the search indices up-to-date. It is executed as follows:
 
     v1Api: forever server/update-search.js --index v1_api_v2 --type item
-    v2Api: forever server/update-search.js --index v2_api_v2 --type item --apiVersion 2
```
matthew-andrews: i think this one is still needed

Procfile
i-like-robots: Maybe we should rename it =]

matthew-andrews: good idea :+1:

matthew-andrews:
i'm going to break thsi into smaller pieces

PR #229, opened 2015-12-14
Translate editorial horizontal rules in to HTML

commuterjoy:
Migrated from next-article

https://github.com/Financial-Times/next-article/blob/master/server/transforms/replace-hrs.js

/cc @matthew-andrews @andygnewman


andygnewman:
:+1:

Issue #219, opened 2015-12-02
Health checks should be inert


matthew-andrews:
Currently they republish missing content.  They shouldn't.  They should only check.  The republish should be a separate process.

PR #230, opened 2015-12-14
Make health check inert and extract latest story updating to process

adgad:
/cc @matthew-andrews (I made up the regions and update interval...)

Addresses https://github.com/Financial-Times/next-es-interface/issues/219


matthew-andrews:
:+1:

PR #231, opened 2015-12-14
Add metadata for fastft items

tom-parker:
@matthew-andrews does this make sense?


server/models/article-v3.js
```diff
@@ -182,7 +182,8 @@ function normalizeItemV2(articleV2) {
 			enabled: articleV2.comments && articleV2.comments.enabled
 		},
 		webUrl: articleV2.webUrl,
-		realtime: articleV2.realtime
+		realtime: articleV2.realtime,
+		brands: articleV2.brands
```
matthew-andrews: this makes sense but i think would prefer not to add the `brands` to the data model until we need it…

i forgot to mention that we actually use `api.ft.com/enrichedcontent` rather than `api.ft.com/content` — and that has the Fast FT tag in the `annotations` part of the data in addition to the `brands` section.

server/models/article-v3.js
```diff
@@ -275,6 +276,15 @@ function applyModel(articleV1, articleV2) {
 	// As a separate field we will also store V2 ‘annotations’
 	article.annotations = metadataV2(articleV2);
 
+	if (article.brands && article.brands.indexOf('http://api.ft.com/things/5c7592a8-1f0c-11e4-b0cb-b2227cce2b54') !== -1 ) {
```
matthew-andrews: so this could change to something like

```js
if (article.annotations && article.annotations.some(tag => tag.idV2 === '5c7592a8-1f0c-11e4-b0cb-b2227cce2b54') {
  article.metadata.push({ /* etc */ });
}
```

server/models/article-v3.js
matthew-andrews: have a look at:-
http://api.ft.com/enrichedcontent/2cced02c-8895-3649-abf2-0da235b08803?apiKey=XXX

matthew-andrews:
Almost perfect — would like to tweak slightly.

matthew-andrews:
:+1:

PR #244, opened 2015-12-17
Update metadata-transformer.js

musthemag:
Added 'Limited' to the list of suffixes to be suppressed


matthew-andrews:
magic :+1:

PR #254, opened 2015-12-21
Migrate TMEs endpoint to a speedy V3 implementation

matthew-andrews:
fyi @wheresrhys


matthew-andrews:
this will bring podcasts into search suggestor… …… without the flag ……… YOLO

server/controllers/tmes.js
```diff
@@ -15,18 +15,40 @@ module.exports = function(req, res, next) {
 			method: 'POST',
 			body: JSON.stringify({
 				size: 0,
+				query: {
+					match: {
+						'metadata.taxonomy': req.params.taxonomy
+					}
+				},
 				aggs: {
-					'primary-sections': {
```
wheresrhys: :+1: I was beginning to think this inaccurate label was immortal

PR #258, opened 2015-12-22
falsey mainImage when missing

matthew-andrews:


matthew-andrews:
Before [this](https://github.com/Financial-Times/next-es-interface/blob/falsey-main-image/server/models/article-v3.js#L310) was true (bad).

@adgad

adgad:
:+1:

PR #267, opened 2015-12-23
add copyright notice to preresolved images

andygnewman:
cc: @matthew-andrews


server/lib/preresolve-images.js
```diff
@@ -42,7 +42,9 @@ function templateMediaResource(mediaResource) {
 	const longdesc = mediaResource.longdesc ? ` alt="${mediaResource.longdesc}"` : '';
 	const width = mediaResource.pixelWidth ? ` width="${mediaResource.pixelWidth}"` : '';
 	const height = mediaResource.pixelHeight ? ` height="${mediaResource.pixelHeight}"` : '';
-	return `<img src="${mediaResource.binaryUrl}"${alt}${longdesc}${width}${height} />`;
+	const copyright = mediaResource.copyright && mediaResource.copyright.notice
+		? ` copyright="${mediaResource.copyright.notice}"` : '';
```
matthew-andrews: whilst the others are valid HTML attributes for images, i don't think that `copyright` is … perhaps `data-copyright-notice` or just `data-copyright`?

matthew-andrews:
one small comment, otherwise :+1:

andygnewman:
Learnt something new there about html attributes - you can only make them up if you put data- before them. Thanks

matthew-andrews:
haha …  umm … kind of.

You can also have custom elements …

http://www.html5rocks.com/en/tutorials/webcomponents/customelements/

Then pretty much anything goes but there has to be a `-` in the tag name …

matthew-andrews:
http://www.w3.org/TR/2011/WD-html5-20110525/elements.html#embedding-custom-non-visible-data-with-the-data-attributes

matthew-andrews:
nice that we could merge this, by the time we come back from after christmas we'll have lots of lovely copyright notices coming through for our articles :-)

andygnewman:
Also got some cleaning up to do post primaryBrand, which should now have worked through the system.

PR #269, opened 2016-01-04
Fix missing longdesc attributes for images.

i-like-robots:
We're currently missing image captions in Next and we may be getting incorrect alt text as the attribute is duplicated.


matthew-andrews:
My bad, sorry :+1:

PR #276, opened 2016-01-19
add opening from V2 to article model

andygnewman:
Add the new opening element created by UP team for wordpress initial content.


matthew-andrews:
:+1:

server/models/article-v3.js
```diff
@@ -221,24 +225,29 @@ function mainImageV2(articleV2) {
 }
 
 function v2Resolutions(articleV2) {
-	let articleLinks = preresolveLinks.findArticleLinks(articleV2.bodyXML);
-	let imageSets = preresolveImages.findImageSets(articleV2.bodyXML);
+	const toResolve = ['bodyXML', 'openingXML'];
+	toResolve.map(item => {
```
i-like-robots: `map` returns a new array, just use `forEach` here

tests/server/models/article-v3-spec.js
```diff
@@ -141,12 +141,17 @@ describe('Article v3 model', function() {
...
 		});
 
-		it('stores V2 body but removes newlines, extraneous whitespace, empty paragraphs and redundant break elements', function() {
+		it('stores V2 body and opening but removes newlines, extraneous whitespace, empty paragraphs and redundant break elements', function() {
 			subject.bodyXML.should.not.match(/\n/);
 			subject.bodyXML.should.not.match(/\s\s/);
 			subject.bodyXML.should.not.match(/<p>\s*<\/p>/);
 			subject.bodyXML.should.not.match(/<\/p><br\/><p>/);
 			subject.bodyXML.should.not.match(/<body>/);
+			subject.openingXML.should.not.match(/\n/);
```
matthew-andrews: how come it's called `opening` in CAPI but `openingXML` in ES?
i think it would be good for us to match capi, even if capi is a little nonsensical.

server/models/article-v3.js
```diff
@@ -221,24 +225,29 @@ function mainImageV2(articleV2) {
 }
 
 function v2Resolutions(articleV2) {
-	let articleLinks = preresolveLinks.findArticleLinks(articleV2.bodyXML);
-	let imageSets = preresolveImages.findImageSets(articleV2.bodyXML);
+	const toResolve = ['bodyXML', 'openingXML'];
+	toResolve.forEach(item => {
```
matthew-andrews: oops we lost the promise here …  it wasn't returned and so this logic no longer blocked :-)

PR #278, opened 2016-01-25
Actually return the promise

tom-parker:


tom-parker:
cc @andygnewman @matthew-andrews

andygnewman:
:+1:

PR #283, opened 2016-01-28
Block resolving the v2Resolutions promise on the logic actually finishing…

matthew-andrews:
:grin:


server/models/article-v3.js
```diff
@@ -225,8 +225,7 @@ function mainImageV2(articleV2) {
 }
 
 function v2Resolutions(articleV2) {
-	const toResolve = ['bodyXML', 'openingXML'];
-	toResolve.forEach(item => {
+	return Promise.all(['bodyXML', 'openingXML'].map(item => {
```
i-like-robots: Should this be filter?

matthew-andrews: It needs both…

server/models/article-v3.js
```diff
@@ -225,32 +225,29 @@ function mainImageV2(articleV2) {
...
-				articleV2[item] = preresolveImages.replaceImageSets(articleV2[item], imageSets, mediaResources);
-				return articleV2;
-			})
-			.catch(function(error) {
-				logger.warn(`Images could not be resolved: ${error.toString()}`);
-				return articleV2;
-			});
-		}
+	return Promise.all(['bodyXML', 'openingXML']
+		.filter(item => return articleV2[item])
```
i-like-robots: return?

matthew-andrews:
Oops :-)
https://github.com/Financial-Times/next-es-interface/pull/276/files#r51110447

PR #285, opened 2016-02-01
rename opening and add test coverage

andygnewman:
@matthew-andrews 

Now that opening has been renamed openingXML in CAPI V2.

Have also done a better test for it.


matthew-andrews:
:+1:

PR #286, opened 2016-02-01
Fix bug where an image is reused it would only get pulled in once

matthew-andrews:


i-like-robots:
I'll admit it, I don't understand why this would behave any differently 😳

i-like-robots:
OH

i-like-robots:
`/g`

i-like-robots:
:ship:

matthew-andrews:
:grin:

PR #301, opened 2016-02-22
Don't fail the check cos of network issues 

aintgoin2goa:
This check is a but flappy and keeps failing cos of network issues with the CAPI.

Have added a new check for the CAPI (at severity 3) and silenced check failures caused by the check itself

This should mean to Matts need to woken unecessarily


matthew-andrews:
hmm, i'm not sure this is right.

matthew-andrews:
i had a chat with matt c earlier.  i don't think we should disable this check, the check itself is fine — it was correctly failing this weekend because of on-going mashery issues.  i think it would be better to either bypass mashery for this check or possibly add retries.

aintgoin2goa:
I guess we don't need this any more if we're just bypassing mashery

PR #303, opened 2016-02-22
remove IPTC and ICB on metadata transform

laurieboyes:
@phamann @matthew-andrews


matthew-andrews:
:sparkles: 

it's really good to have more people knowing how this works

PR #304, opened 2016-02-22
only fail check if we get three errors in a row

aintgoin2goa:


server/checks/latest-news.js
```diff
@@ -6,6 +6,8 @@ const checkLatestContent = require('../lib/check-latest-content');
 module.exports = () => {
 	const region = process.env.REGION === 'US' ? 'US' : 'EU';
 	const interval = 30*1000;
+	let errorThreshold = 3;
```
matthew-andrews: const?

server/checks/latest-news.js
```diff
@@ -31,8 +34,9 @@ module.exports = () => {
 				logger.info(`esindex=3 region=${region} event=LATEST_NEWS_CHECK_RUN missing=${missing}`);
 			})
 			.catch(err => {
+				errorCount++;
 				lastUpdated = new Date();
-				ok = false;
+				ok = (errorCount >= errorThreshold);
 				checkOutput = 'This cluster was unable to check if the latest news has been successful published.';
```
matthew-andrews: can we add `… for $(errorCount) times in a row`

matthew-andrews:
:+1:

PR #312, opened 2016-03-02
Temporarily map subjects to other taxonomies before we ditch subjects completely

laurieboyes:
@phamann @matthew-andrews 
Decisions I'm happy for a second opinion on:
- As it's temporary we're better hard-coding the mappings then messing around talking to bertha directly
- Only written a test for a single subject, as they're all the same in principal


metadata/temp-subject-mappings.js
```diff
@@ -0,0 +1,834 @@
+'use strict';
+
+//// https://bertha.ig.ft.com/view/publish/gss/1moVePfkBMjDvL0TUmzkatmbHWjSxtVrmiXqWDNr9xwQ/Sheet1
```
phamann: I'd leave a comment above this as to why it's here, as currently looks like accidental debugging.

phamann:
:+1:

matthew-andrews:
:+1: go go go

PR #316, opened 2016-03-03
Revert "Temporarily switch back to Mashery"

matthew-andrews:
Reverts Financial-Times/next-es-interface#315


matthew-andrews:
Ready for the next test

PR #317, opened 2016-03-04
Get correct image size when not already present, from image service metadata endpoint

andygnewman:
Objective:

Get image sizes for any image that doesn't have them already set.

Article - in process of pre-resolving images
Podcasts - in setting the main image

For square images (especially seen in podcasts) - constrains square image to max 394 (which is the height of a 700px image with a 16:9 aspect ratio.

Has transformed the podcast model into a promise, so should be checked carefully.


server/lib/get-image-size.js
```diff
@@ -0,0 +1,31 @@
...
+	if (image) {
+		const imageUrl = encodeURIComponent(image);
+		const url = `https://next-geebee.ft.com/image/v1/images/metadata/${imageUrl}?source=next`;
+		return fetch(url, {timeout: 2000})
+		.then(fetchres.json)
+		.then(response => {
+			let imageWidth = response.width;
+			let imageHeight = response.height;
+			// HACK resize square images to have height of a standard 16:9 aspect ratio on width 700
+			if (imageWidth === imageHeight && imageWidth > 394) {
```
andygnewman: remove this and look at image resize logic with max height

server/lib/get-image-size.js
```diff
@@ -0,0 +1,31 @@
+'use strict';
+
+const fetchres = require('fetchres');
+const logger = require('ft-next-logger').logger;
+
+module.exports = (image) => {
+	if (image) {
```
andygnewman: remove this from here just don't call the function if no image

server/lib/preresolve-images.js
```diff
@@ -18,6 +19,19 @@ function createRequest(url, articleUuid) {
...
 		})
+		.then(imageJson => {
+			if (!imageJson.pixelWidth || !imageJson.pixelHeight) {
+				return getImageSize(imageJson.binaryUrl)
+				.then(imageSize => {
+					if (imageSize) {
+						imageJson.pixelWidth = imageSize.imageWidth;
+						imageJson.pixelHeight = imageSize.imageHeight;
+					}
+					return imageJson;
```
andygnewman: `imageSize && Object.assign(imageJson, imageSize)`

server/models/podcast-v3.js
```diff
@@ -10,6 +11,8 @@ module.exports = function podcastModelV3(data, feedMetadata) {
 	let image = data['itunes:image'].href || feedMetadata.defaultImage;
 	let primarySectionId = metadataV1.primarySection.term.id;
 	let tagList = [];
+	let imageWidth = 394;
```
andygnewman: get rid of the defaults

server/models/podcast-v3.js
```diff
@@ -21,35 +24,43 @@ module.exports = function podcastModelV3(data, feedMetadata) {
...
-			feedMetadata.masterSource
-		],
-		attachments: [
-			{
-				mediaType: data.enclosure.type,
-				url: data.enclosure.url
+	return getImageSize(image)
+		.then(response => {
+			if (response) {
+				imageWidth = response.imageWidth;
```
andygnewman: don't do this put them straight into main image below

server/lib/preresolve-images.js
```diff
@@ -18,6 +19,15 @@ function createRequest(url, articleUuid) {
 			const url = data.members[0].id;
 			return fetchCapiJson(url);
 		})
+		.then(imageJson => {
+			if (!imageJson.pixelWidth || !imageJson.pixelHeight) {
+				return getImageSize(imageJson.binaryUrl)
+				.then(imageSize => {
+					return imageSize && Object.assign(imageJson, imageSize);
```
i-like-robots: if `imageSize` does not come back this will return a blank. My bad. Should be...

```
return imageSize ? Object.assign(imageJson, imageSize) : imageJson
```

ironsidevsquincy: ```.then((imageSize = {}) => Object.assign(imageJson, imageSize))```

i-like-robots: App is still Node 4 (without transpilification)

PR #319, opened 2016-03-09
Remove fastly logging, it's broken and causes Fastly issues

matthew-andrews:


i-like-robots:
I am totally confident that this is reasonable varnishness.

PR #325, opened 2016-03-21
move xslt transforms into ingester app

andygnewman:
To Do

- [x] Move xslt transforms from article app
- [x] Make sure unit tests for each transform still work
- [x] Add promise into article v3 controller based on bodyXML and openingXML
- [x] Add to overall article model as bodyHTML and openingHTML
- [x] Write integration tests for bodyHTML and openingHTML


server/models/article-v3.js
```diff
@@ -266,6 +267,13 @@ function cleanAndSortStrings (items) {
 		});
 }
 
+function contentHtmlTransform (contentXml, params) {
+	if (contentXml) {
+		return contentXsltTransform(contentXml, 'main', params);
+	}
+	return Promise.resolve();
```
i-like-robots: Just don't call this method?

server/models/article-v3.js
```diff
@@ -326,10 +334,27 @@ function applyModel (articleV1, articleV2) {
...
+	let bodyHTML = contentHtmlTransform(article.bodyXML, xsltParams).then(transformedHtml => {
+		if (transformedHtml) {
+			article.bodyHTML = transformedHtml;
+		}
+	});
+
+	let openingHTML = contentHtmlTransform(article.openingXML, xsltParams).then(transformedHtml => {
+		if (transformedHtml) {
+			article.openingHTML = transformedHtml;
+		}
```
i-like-robots: I think the condition is a bit pointless, `article.openingHTML = undefined` won't do any harm.

```js
JSON.stringify({ param: undefined }) // => {}
```

server/models/article-v3.js
```diff
@@ -326,10 +334,27 @@ function applyModel (articleV1, articleV2) {
 		}
 	});
 
+	let xsltParams = {
+		id: article.id,
+		webUrl: article.webUrl
+	};
+
+	let bodyHTML = contentHtmlTransform(article.bodyXML, xsltParams).then(transformedHtml => {
```
i-like-robots: `const bodyHTML = article.bodyXML && contentHtmlTransform(...)` will do (see above)

tests/server/stylesheets/transform-helper.js
```diff
@@ -0,0 +1,8 @@
+'use strict';
+
+var contentXSLT = require('../../../server/stylesheets/content-xslt');
+
+module.exports = function(xml) {
+
+	return contentXSLT(xml, 'main');
+};
```
i-like-robots: This helper isn't much help now?

i-like-robots:
Awesome

andygnewman:
Before we merge, do we need to do something to the build process for the libxslt? as with article?

i-like-robots:
The XSLT tests are all green so it would appear not

andygnewman:
That's a good point.

matthew-andrews:
does this not require an update to the V3 elastic search index model?

matthew-andrews:
^^ to answer my question, the answer is no :-)
https://github.com/Financial-Times/next-es-interface/blob/master/schema/content.json#L37-L48

andygnewman:
Thanks to @i-like-robots for adding the new fields in during the reindexing.

andygnewman:
@matthew-andrews can you help me understand what went wrong with this merge?

matthew-andrews:
i'll pop over in a bit

PR #330, opened 2016-03-22
Sanity check on ingest

i-like-robots:
screen shot 2016-03-22 at 15 57 38

This will check that a processed piece of data is not committing heinous crimes against content.

(Also deletes the old article model)


i-like-robots:
Fixes https://github.com/Financial-Times/next-es-interface/issues/324

matthew-andrews:
looks sane to me, similar to what we had on the web app ^^

PR #333, opened 2016-03-23
Add initialPublishedDate

ironsidevsquincy:
@andygnewman


matthew-andrews:
wait a tick, what's this for? have we considered asking UPP to provide this in v2?

tests/server/models/article-v3-spec.js
```diff
@@ -422,10 +423,14 @@ describe('Article v3 model', function() {
 				.then(response => subject = response);
 		});
 
-		it('should prefer publishedDate\'s from CAPI v1', function() {
+		it('should prefer publishedDate\'s from CAPI v1 because v2 is sometimes completely wrong', function() {
```
matthew-andrews: this bug is actually fixed now, so we can go back to using `publishedDate` from v2 :-) // cc @andygnewman

ironsidevsquincy:
For new/updated labels on cards, i.e was this article published in the last hour (new)/was in updated in the last hour (updated)

Is that the process? Thought we were using our own ES? Can we ask for this to be added to v2 but use the v1 data for now?

matthew-andrews:
1. i think this change requires an Elasticsearch schema update+release.  https://github.com/Financial-Times/next-es-interface/blob/master/schema/content.json#L52-L55
2. es-interface's data model should be matching UPP's future format wherever possible so we should at least ask them ‘when you come to implement this feature, what will it look like’.
3. it's fine to add this here for now as you described, but in the future it would be good to first give the content api team a chance to implement this before hacking around it via ES.

ironsidevsquincy:
Is this hacking around? Even if it was in v2 we're still going through ES for everything, so this would still need to be done (just getting the data internally from v2 instead of v1)? What am I missing?

Ok, need to make sure the field name matches up with UPP's future field. Who should I talk to?

matthew-andrews:
any additional dependency on v1 APIs is kind is kind of a hack…

have a chat with @guyvalerio or @martingartonft or @davidlintonattheft about whether `initialPublishedDate` in capi v2 will be called `initialPublishedDate`

i-like-robots:
Seems like an editorial issue... if `publishedDate` != actual publication date then I feel like that's not our problem.

matthew-andrews:
mmmmm it's not quite that simplistic matt.  there's the date the article was first published and the date the article was last published.  both are interesting/valid…

ironsidevsquincy:
Yep, the `publishedDate` is actually the *last* published date, which is a bit confusing, that name to me implies the first time it was published, but obviously can't change that now, so to get the actual published date I've added a new field `initialPublishedDate`. Whether this is the right name, that's up for debate

i-like-robots:
OK I see the value in both... and that the V2 property is ambiguous.

matthew-andrews:
grab a @guyvalerio, @martingartonft or @davidlintonattheft and ask them what it should be called ^^

martingartonft:
initialPublishedDate is fine I think.

matthew-andrews:
Cool — @i-like-robots can you show mr hurley how to update the schema/do it for him?

i-like-robots:
Updating the schema = new index... since this is only for a display property it should be OK for now.

matthew-andrews:
even for adding new property? :scream:

i-like-robots:
Can't update the settings (inc. mapping) for an open index, we must close it first which we can't do on AWS.

i-like-robots:
New properties will be automagically added to the mapping but it's not perfect =]

matthew-andrews:
ok cool +1 from me

ironsidevsquincy:
Thanks all

PR #334, opened 2016-03-23
Migrate article transforms

andygnewman:
Move some of the transforms from article to ingest article v3 model
https://github.com/Financial-Times/next-article/blob/master/server/transforms/body.js#L33

Namely
- replace ellipses
- remove spaces after links before punctuation
- trimmed links


i-like-robots:
:+1: Looks great. I'm doing some refactoring of these modules so will help with that too =)

andygnewman:
Closed in favour of this one https://github.com/Financial-Times/next-es-interface/pull/340 that is on the refactored code baseline.

PR #336, opened 2016-03-23
Temporarily remove XSLT HTML transform while we work out why it's broken

i-like-robots:
=(

Got lots of unresolved `<ft-content>` elements which were being ingested, now the sanity check chucks them out.

Easier to delete than revert the revert of the revert revert.


i-like-robots:
CC @andygnewman @matthew-andrews

PR #339, opened 2016-03-24
Article model refactor stage 1

i-like-robots:
Before this PR ends up huge and unreadable...

Simply extracts out a bunch of functionality into standalone modules to reduce noise and encourage more focussed test coverage.

Merge after https://github.com/Financial-Times/next-es-interface/pull/334


andygnewman:
Looking good. I'll hold off merging my additional transforms in, until after this one has gone in as one of us will need to rebase and it's probably easier if I do it on to this than the other way round.

PR #340, opened 2016-03-24
add xml transforms from article

andygnewman:
Does what this PR https://github.com/Financial-Times/next-es-interface/pull/334 did, but on the refactored code.


i-like-robots:
+1 but I don't fancy merging at 5.20pm before a bank holiday!

andygnewman:
Agree, will do it on Tuesday.

PR #341, opened 2016-03-24
Article model refactor second act

i-like-robots:
Refactors the link and image resolution modules so that their interface becomes a single method making their integration far more simple to reason about.

Found [a problem](https://github.com/Financial-Times/next-es-interface/issues/342) when trying to clean up nock mocks so there's a little extra noise in the tests to scope them.


andygnewman:
Looking good, just those unused vars to deal with, but certainly becoming more easily understood.

PR #343, opened 2016-03-29
Update search updater spec 

### LINKED ISSUE #342: Update spec has never worked
The expectations are never run in this spec:

https://github.com/Financial-Times/next-es-interface/blob/master/tests/update-spec.js

When switching to run async (I.E. with `done()` callback) they fail.

Also prevents using `nock` module unscoped.

i-like-robots:
Closes #342 (this spec did sod all)

It was also diving deeply into other modules ಠ_ಠ (update calls article model which calls a billion things).


tests/server/search-updater-spec.js
```diff
@@ -0,0 +1,65 @@
...
+const SearchUpdater = proxyquire('../../server/search-updater.js', {
+	'./lib/api-v3': apiV3Mock
+});
+
+var since = '2015-01-07T16:24:11.508Z';
+var nock = require('nock');
+var notificationsPage1 = require('../fixtures/notificationsPage1.json');
+var notificationsPage2 = require('../fixtures/notificationsPage2.json');
+var notificationsPage3 = require('../fixtures/notificationsPage3.json');
+var apiResponse = require('../fixtures/apiResponse.json');
```
andygnewman: unused var

andygnewman:
LGTM

PR #345, opened 2016-03-30
Move nock mock setup into before block so that they don't get cleaned…

i-like-robots:
… up by any other tests. Tests run in a random order so this didn't come up before.


andygnewman:
:+1:

PR #346, opened 2016-03-30
Reinstate html transforms

i-like-robots:
Reinstates the XSLT XML to HTML transforms after some refactoring.

CC @andygnewman


server/models/article-v3.js
```diff
@@ -123,8 +116,32 @@ function applyModel (articleV1, articleV2) {
...
+			.then(result => {
+				article.openingXML = result;
+				return xslt(article.openingXML, 'main');
+			})
+			.then(result => {
+				article.openingHTML = result;
+			})
+	);
+
+	resolutions = Promise.all(resolutions);
```
i-like-robots: I tried to make this as readable as possible, not DRY. Ideas for more readability welcome!

server/models/article/body-v1.js
```diff
@@ -12,8 +12,12 @@ function isSlideshow (article) {
 
 module.exports = (article) => {
 	let body = (article.item.body && article.item.body.body || '')
+		// Remove new lines
+		.replace(/\r?\n|\r/g, '')
+		// Remove extraneous whitespace
+		.replace(/\s{2,}/g, ' ')
 		// Remove all tags except paragraphs
-		.replace(/<(?!\s*\/?\s*p\b)[^>]*>/gi, '');
+		.replace(/<(?!\s*\/?\s*p\b)[^>]*>/gi, '')
```
i-like-robots: Poor linting effort there

server/models/article-v3.js
andygnewman: You could do 
```
if (articleV2) {
    let resolutions = [];

    article.bodyXML && ...
    
    article.openingXML && ...

    resolutions = Promise.all(resolutions);
}
```

but doesn't really add much

matthew-andrews: can be `const resolutions` 🙃

i-like-robots: whut?

server/models/article-v3.js
```diff
@@ -123,25 +111,53 @@ function applyModel (articleV1, articleV2) {
...
+							title: image.title,
+							description: image.description,
+							url: image.binaryUrl,
+							width: image.pixelWidth,
+							height: image.pixelHeight
+						};
+					}
+				})
+		);
+	}
```
i-like-robots: I think this block can be cleaned up if we change the handling of `mainImage` but that's for another day =]

andygnewman:
:+1:

PR #348, opened 2016-03-31
We don't currently care about V2 annotations so don't check their sanity

i-like-robots:
I'm not sure this makes much difference but the point is, annotations aren't any use to us right now.


i-like-robots:
Actually, I realise if content is V2 only this will :boom:

i-like-robots:
Then again if content is V2 only it probably won't show up on the site.

matthew-andrews:
yeah i think that's fine, we should delete the old fastFT posts first and then do a quick search to see what's left that would be affected + make a decision on that before we merge

matthew-andrews:
i think this is still ok, matt — deleting all the v2 only stuff now :grin:

matthew-andrews:
we should merge this soon
http://next-elastic.ft.com/v3_api_v2/item/_search?q=_missing_:metadata

PR #349, opened 2016-04-01
Add standout property

ironsidevsquincy:
Now coming through v2

screen shot 2016-04-01 at 11 24 04

@matthew-andrews


matthew-andrews:
lgtm

PR #352, opened 2016-04-05
Filter dodgy articles

matthew-andrews:


server/lib/sanity-check.js
```diff
@@ -4,6 +4,7 @@ const uuid = /[a-f0-9]{8}-[a-f0-9]{4}-[a-f0-9]{4}-[a-f0-9]{4}-[a-f0-9]{12}/;
 const empty = /^\s+$/;
 // const ftcontent = /\<\/?ft-content/;
 
+// COMPLEX: This *returns* true OR Error objects.  Be warned…
```
matthew-andrews: @i-like-robots this was super confusing…  my humble opinion that it would be more natural for this function to `throw` errors and consistently return a boolean rather than return both.

andygnewman:
LGTM

server/lib/sanity-check.js
i-like-robots: Sure, it originally only returned booleans but the logging wasn't useful that way

PR #356, opened 2016-04-07
Up continuous updater check threshold to 10

phamann:
ronseal


server/checks/content-updating.js
```diff
@@ -44,7 +44,7 @@ module.exports = () => {
 				if (unique(oldest).length > 1) {
 					ok = true;
 					checkOutput = "All's well" + debug;
-				} else if (oldest.length > 4) {
+				} else if (oldest.length > 10) {
```
matthew-andrews: also need to update line 41…

PR #357, opened 2016-04-08
Make sanity check always return one type of value

i-like-robots:
Because checking the type is lame. `if (sanityCheck) { .. }` :scream:


matthew-andrews:
thanks ! <3

PR #358, opened 2016-04-08
get rid of empty elements

andygnewman:
cc: @i-like-robots 

Get better at removing empty elements.

eg. nested empty em, span, strong before removing the enclosing p
eg. empty p's left after removing dodgy interactive graphics


server/models/article/xml-v2.js
```diff
@@ -10,6 +10,12 @@ module.exports = (xml) => {
 		.replace(/\s{2,}/g, ' ')
 		// Remove spaces between end of links and punctuation
 		.replace(/<\/a>\s+([,;.:])/g, '</a>$1')
+		// Remove empty em tags
+		.replace(/<em>\s*<\/em>/g, '')
+		// Remove empty strong tags
+		.replace(/<strong>\s*<\/strong>/g, '')
+		// Remove empty span tags
+		.replace(/<span>\s*<\/span>/g, '')
```
i-like-robots: Should we include `<b>` too?

server/models/article/xslt.js
```diff
@@ -46,7 +46,10 @@ module.exports = function(xml, stylesheet, params) {
 				return reject('xsltproc exited with code ' + code + ': ' + errors);
 			}
 
-			resolve(output.join('').replace(/<\/?(html|body)>/g, ''));
+			resolve(output.join('')
+				.replace(/<\/?(html|body)>/g, '')
+				.replace(/<p><\/p>/g, '')
```
i-like-robots: Should we solve this in the XSLT?

tests/server/models/article/xml-v2-spec.js
```diff
@@ -25,6 +25,26 @@ describe('XML V2', () => {
...
+		const xmlContent = "<p>A <em></em> Para</p>";
+		expect(subject(xmlContent)).to.not.match(/<em><\/em>/);
+	});
+
+	it('removes empty strong tags', () => {
+		const xmlContent = "<p>A <strong></strong> Para</p>";
+		expect(subject(xmlContent)).to.not.match(/<strong><\/strong>/);
+	});
+
+	it('removes empty sspan tags', () => {
```
i-like-robots: typo

tests/server/models/article/xml-v2-spec.js
```diff
@@ -25,6 +25,26 @@ describe('XML V2', () => {
...
+		const xmlContent = "<p>A <span></span> Para</p>";
+		expect(subject(xmlContent)).to.not.match(/<span><\/span>/);
+	});
+
+	it('removes <p>\'s with enclosed empty elements', () => {
+		const xmlContent = "<p>one</p><p><em></em></p><p>two</p>";
+		expect(subject(xmlContent)).to.not.match(/<p><em><\/em><\/p>/);
```
i-like-robots: Would probably add `expect(subject(xmlContent)).to.not.match(/<p><\/p>/)` as well

server/models/article/xslt.js
andygnewman: That was my first instinct also.
However the situation occurs when we remove an <a> because it doesn't meet certain conditions. This could happen whether or not it it is the only thing in a <p> which makes the logic for removing it quite complex and doing a pure match on an empty <p> won't work because the empty <p> is the result of a different transformation that could happen in any order.

So this felt like the simplest and most scalable (there may be other scenarios that we haven't discovered yet) solution. (Though any simple XSLT suggestions welcomed)

i-like-robots: Do you know which rule it is that creates the empty para? We should be able to say 'only create this p if it has children of any kind'.

server/models/article/xml-v2.js
```diff
@@ -10,6 +10,14 @@ module.exports = (xml) => {
...
 		// Remove spaces between end of links and punctuation
 		.replace(/<\/a>\s+([,;.:])/g, '</a>$1')
+		// Remove empty em tags
+		.replace(/<em>\s*<\/em>/g, '')
+		// Remove empty b tags
+		.replace(/<b>\s*<\/b>/g, '')
+		// Remove empty strong tags
+		.replace(/<strong>\s*<\/strong>/g, '')
+		// Remove empty span tags
+		.replace(/<span>\s*<\/span>/g, '')
```
i-like-robots: Just for the record I tried this with backrefs (`<(p|em|b|strong)>\s*<\/\1>`) and of course it will fail when nested, E.G.

```
'<p><em></em></p>.replace(/<(p|em|b|strong)>\s*<\/\1>/, '') // <p><p>
```

This approach suffers from the same issue but hopefully less so.

server/stylesheets/interactive-graphics.xsl
```diff
@@ -1,7 +1,13 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <xsl:stylesheet xmlns:xsl="http://www.w3.org/1999/XSL/Transform" version="1.0">
 
-    <xsl:template match="a[@data-asset-type='interactive-graphic' and contains(@href, 'https')]">
+    <xsl:template match="p[a[@data-asset-type='interactive-graphic']]">
+      <xsl:if test="current()[a[contains(@href, 'https')]]">
+        <xsl:apply-templates select="a[@data-asset-type='interactive-graphic' and contains(@href, 'https')]" mode="interactive-graphic" />
+      </xsl:if>
+    </xsl:template>
```
i-like-robots: do we still need to match the link not in a paragraph?

```xslt
<xsl:template match="p[a[@data-asset-type='interactive-graphic']]">
  <xsl:apply-templates select="a[@data-asset-type='interactive-graphic' and contains(@href, 'https')]" />
</xsl:template>

<!-- maintain the existing --->
<xsl:template match="a[@data-asset-type='interactive-graphic' and contains(@href, 'https')]">
 ...
</xsl:template>
```

server/models/article/xml-v2.js
andygnewman: Thanks. I think with the ordering I have that it should be reasonably robust for nested.

server/stylesheets/interactive-graphics.xsl
andygnewman: All our original tests cases had them within p's, but have added in tests where not in a p. Should also do a check that there is no other content in the p except the IG before getting rid of the p I guess.

i-like-robots:
🚢

PR #359, opened 2016-04-11
pass tweet attributes through on blockquotes

andygnewman:
cc: @georgecrawford @i-like-robots 

For blockquotes;
- passes through existing classes (eg. twitter-tweet)
- adds id of tweet as data-tweet-id attribute


georgecrawford:
Thanks Andy. LGTM.

tests/server/stylesheets/blockquotes-spec.js
```diff
@@ -16,11 +16,28 @@ describe('Blockquotes', function () {
 			, 'main')
 			.then(function (transformedXml) {
 				transformedXml.should.equal(
-					'<blockquote class="n-content-blockquote">' +
+					'<blockquote class="n-content-blockquote ">' +
```
i-like-robots: a good example for switching to HTML assertions rather than string matching ;-]

roland-vachter:
+1

i-like-robots:
So post discussion 6 months later I guess we're going for appending the class name `n-content-blockquote--tweet` if the original `<blockquote>` has the class name `twitter-tweet`. That sound right?

andygnewman:
Yep, have retained the original twitter-tweet class aswell in case we every use the twitter widget to embed the official rendering as per https://dev.twitter.com/web/embedded-tweets.

PR #363, opened 2016-04-13
remove newline on xslt output and handle non-video links with no content

andygnewman:
- removes unnecessary newline character from xslt output
- removes any link that doesn't have content (except when has a data-asset-type - eg. video)
- better handles <p>'s with only whitespace left behind when images moved out and other stuff removed


server/stylesheets/external-image.xsl
```diff
@@ -11,14 +11,14 @@
...
-        <xsl:if test="count(child::node()[not(self::img)]) &gt; 0">
+        <xsl:if test="count(child::node()[not(self::img)]) &gt; 0 and not(normalize-space(current()/text()) = '')">
             <p><xsl:apply-templates select="child::node()[not(self::img)]" /></p>
         </xsl:if>
     </xsl:template>
 
     <xsl:template match="p[a/img]">
         <xsl:apply-templates select="a/img" mode="figure-content" />
-        <xsl:if test="count(child::node()[not(self::a)]) &gt; 0">
+        <xsl:if test="count(child::node()[not(self::a)]) &gt; 0 and not(normalize-space(current()/text()) = '')">
```
i-like-robots: Can `and not(normalize-space(current()/text()) = '')` be expressed as `and normalize-space(current()/text()) != ''` so it's more familiar? I don't mind just took me a few seconds to figure out!

tests/server/stylesheets/big-number-spec.js
```diff
@@ -20,7 +20,7 @@ describe('Big Number', function () {
 					'<div class="n-content-big-number">' +
 						'<span class="n-content-big-number__title">33m</span>' +
 						'<span class="n-content-big-number__content">These are powerful but fragile emissaries of a culture that not even their descendants remember</span>' +
-					'</div>\n'
+					'</div>'
```
i-like-robots: HTML assertions would avoid this... not that I'm suggesting we re-format these tests!

server/stylesheets/external-image.xsl
andygnewman: Good shout.

PR #364, opened 2016-04-13
Update EC2 tags to new FT-recommended formats

rowanbeentje:
The platform team migrated tags to a new format on March 21st, including updating tags on existing instances; ensure new instances will get set up with the new standard for tags.


matthew-andrews:
lol thanks

PR #366, opened 2016-04-13
Cachebust

georgecrawford:


server/lib/fetch-capi-json.js
```diff
@@ -1,6 +1,15 @@
...
 
+const url = require('url');
+
 module.exports = (endpoint) => {
+
+	// Add cachebust to all CAPI requests, as intermediate caches return stale content despite an update notification
+	const parsed = url.parse(endpoint, true);
+	delete parsed.search;
+	parsed.query.cb = Date.now();
+	endpoint = url.format(parsed);
```
matthew-andrews: thank you for all your effort here, but would you mind restricting this just to the CAPI v1 requests?

here:-
https://github.com/Financial-Times/next-es-interface/blob/master/server/models/article-v3.js#L18

georgecrawford: I had a feeling you might say that. Better now?

tests/server/models/article-v3-spec.js
```diff
@@ -22,6 +22,31 @@ const ARTICLE_ID = 'cf750772-6c4d-11e5-8171-ba1968cf791a';
...
+			})
+			.reply(200, fixtureV1);
+
+		const requestV2 = nock(CAPI_V2_URL)
+			.get(`/${ARTICLE_ID}`)
+			.times(1)
+			.reply(200, fixtureV2);
+
+		return instance(ARTICLE_ID)
+			.then(response => {
```
matthew-andrews: response is defined by not used :] (linting)

matthew-andrews:
i think circle has helpfully found a linting error but otherwise lovey +1

georgecrawford:
Yeah - I've only now discovered `make install` to add the `.eslintrc.json` file.

matthew-andrews:
quick quick merge merge

PR #367, opened 2016-04-15
Update sanity check usage to align with #357

i-like-robots:
Broken since https://github.com/Financial-Times/next-es-interface/pull/357 sorry sorrryyyy


matthew-andrews:
s'ok +1

PR #372, opened 2016-04-18
Actually make shielding work

matthew-andrews:
@aintgoin2goa @phamann


src/backends/prod.json
```diff
@@ -50,5 +54,19 @@
       "expected_response": 200,
       "interval": 1000
     }
+  ],
+  "conditions": [
```
matthew-andrews: actually need to make this work in fastly-tools

matthew-andrews: actually just went ahead and did that

PR #373, opened 2016-04-19
run all the unit tests

andygnewman:
cc: @matthew-andrews Remove the only which means it's only running a single unit test.


matthew-andrews:
oops!

matthew-andrews:
+1

PR #374, opened 2016-04-19
Big number transform

andygnewman:
cc: @i-like-robots 

To bring big number in line with other elements.
- enclose in an `<aside>`
- put associated text in a `<footer>`
- unwrap first level `<p>` encapsulating the number or associated text, whilst retaining any other html


i-like-robots:
Closing as it's been inactive a long time

PR #381, opened 2016-04-21
VCL hash by default uses the host header, which we change because of …

matthew-andrews:
…Heroku.  We want to use the customer facing host header so use that for `vcl_hash` rather than the one we send to backends…

There are probably other ways of achieving this.

https://docs.fastly.com/guides/vcl/manipulating-the-cache-key


matthew-andrews:
Before:-
purge-before

After:-
purge

wheresrhys:
oooh. Will something like this need adding to every cdn where we _may_ use shielding?

matthew-andrews:
maybe this was a better idea

PR #382, opened 2016-04-21
Remove photo diary hack that puts the first image as a master image now they're going to do it properly

matthew-andrews:
@andygnewman @musthemag


andygnewman:
👍

matthew-andrews:
will give @musthemag a few days before we merge this 😁

PR #384, opened 2016-04-22
Set the Host only when it needs to be set (between vcl_miss and vcl_fetch)

matthew-andrews:
Same effect as #381, but arguably better.

@richard-still-ft @commuterjoy @wheresrhys @phamann


matthew-andrews:
@wheresrhys i think something like *this* will need to be done on every CDN where we're manipulating the host header in VCL and want to do dynamic purging.

PR #385, opened 2016-04-22
Revert "Set the Host only when it needs to be set (between vcl_miss and vcl_fetch)"

matthew-andrews:
Reverts Financial-Times/next-es-interface#384


matthew-andrews:
so that broke stuff ……

PR #390, opened 2016-04-27
Push api

aintgoin2goa:


server/processes/push-updater.js
```diff
@@ -0,0 +1,87 @@
+'use strict';
+const EventSource = require('eventsource');
+const v3Api = require('../lib/api-v3');
+const log = require('@financial-times/n-logger').default;
+
+// we don't use next-express but code expects GLOBAL.fetch to exist
+// todo: I really hate doing this, think of a way to avoid it
+require('isomorphic-fetch');
```
matthew-andrews: i'm sorry about this

matthew-andrews:
looks good to me, do you concur @i-like-robots ?

integration-tests/push-api.int.js
```diff
@@ -0,0 +1,27 @@
...
+	before(() => {
+		process = require('../server/processes/push-updater');
+	});
+
+	it('Should listen for push events and pass them onto the api - watch the logs to see if it works!', () => {
+		process.start();
+		return wait(TIMEOUT)
+			.then(() => {
+				return true;
+			});
```
i-like-robots: This test doesn't do anything?

i-like-robots: (There are no assertions made)

server/processes/push-updater.js
```diff
@@ -0,0 +1,87 @@
...
+}
+
+function stop(){
+	eventSource.close();
+}
+
+module.exports = {
+	start: start,
+	stop: stop
+};
```
i-like-robots: module.exports = { start, stop } is fine =]

tests/server/processes/push-updater-spec.js
```diff
@@ -0,0 +1,54 @@
...
+const EventSourceStub = {
+	onmessage : function(){},
+	onerror : function(){}
+};
+
+EventSourceStub.__fakeEvent = function(data){
+	this.onmessage({data:JSON.stringify(data)});
+};
+
+const EventSourceMock = sinon.stub().returns(EventSourceStub);
```
i-like-robots: May be confusing to call it `Mock` - because it has no assertions, it's just a stub.

integration-tests/push-api.int.js
aintgoin2goa: no, it doesn't run either - but it is useful to check that UPP are actually implementing the spec (they weren't at first)

tests/server/processes/push-updater-spec.js
aintgoin2goa: true, but EventSourceStub was already taken!  Suggestions?

.gitignore
```diff
@@ -3,7 +3,7 @@
 /public/
 .env
 /.haikro-cache
-.eslintrc.js
+.eslintrc.json
```
matthew-andrews: believe this should be `*.js` for new versions of `n-makefile`

integration-tests/push-api.int.js
i-like-robots: OK, no point leaving unused code around then =]

tests/server/processes/push-updater-spec.js
i-like-robots: Maybe just switch the assertion around, so instead of `sinon.assert.called(EventSourceMock);` set the assertion here (`EventSourceMock.atLeast(1)`) and change line 39 to `EventSourceMock.verify()`

integration-tests/push-api.int.js
aintgoin2goa: I want to keep this - it will come in handy if (when) the api breaks.  I've done this a lot in the past and it's always been useful

i-like-robots: wiki it?

matthew-andrews:
💥

Issue #387, opened 2016-04-23
Ignore some V1 metadata taxonomies


i-like-robots:
We never expose some of the tag data we store in our apps, we have code to ignore the same things in multiple places. This is the full list of V1 metadata taxonomies:

- authors
- brand
- genre
- icb
- iptc
- mediaType
- organisations
- people
- regions
- sections
- specialReports
- subjects (never used?)
- topics

I propose we simply do not store the tags given the icb, iptc and subjects taxonomies.

I believe this will affect < 10 content items currently in our store.

matthew-andrews:
i thought we already moved icb, iptc and subject taxonomies…… see here:-
https://github.com/Financial-Times/next-es-interface/blob/master/server/lib/metadata-transformer.js#L85-L108

i-like-robots:
The tag extractor module uses [these predicates](https://github.com/Financial-Times/next-es-interface/blob/master/server/lib/predicates.js) to pluck out tags including the aforementioned. Although they're present on only a very small number of documents they're labels (beginning with 0) mean they bubble up when sorting alphanumerically.

i-like-robots:
Ahh, the logic you point out @matthew-andrews was added [_after_](https://github.com/Financial-Times/next-es-interface/commit/31b61b9cb473016ec95d2f1cecb7c38ff08a19a6) the new content model was implemented... and updates merge rather than replace. Will change it to `null`.

i-like-robots:
Ignore the above... some articles are not being updated because they're tagged as podcasts and https://github.com/Financial-Times/next-es-interface/blob/master/server/continuous-updater.js#L19

i-like-robots:
content tagged with ICB, IPTC and subjects tags have now been cleared up.

PR #392, opened 2016-05-03
Give icb, iptc and subject metadata taxonomies null values

i-like-robots:
Content exists within the index that predates the implementation of these checks. As ES documents are merged rather than replaced the rogue data remains.

CC #387


i-like-robots:
Nope

matthew-andrews:
;)

Issue #383, opened 2016-04-22
Improve consistency of article HTML


andygnewman:
To benefit from easier styling, less reliance on class names, it is suggested that we clean up our article HTML. Some suggested usages (changes from existing) are below;

- [ ] Headings
`<h1>` Prime Article Headline
`<h2>` In Article Subhead
`<h3>` Title of complementary section / aside
`<h4>` Headline of linked article in an aside

- [ ] Pullquote - should be a blockquote within an aside
```
<aside>
    <blockquote>
        <p>Quote here</p>
        <footer>Here's who said it</footer>
    </blockquote>
</aside>
```
- [ ] Big Number should be in an aside and accompanying text should be a footer (consistent with blockquote) [PR](https://github.com/Financial-Times/next-es-interface/pull/374)
```
<aside>
    <div>22m</div>
   <footer>This tells us what the number represents</footer>
</aside>
```
- [ ] Aside Titles Should Be H3 wrapped in a div (placeholder) [PR](https://github.com/Financial-Times/next-es-interface/pull/376)
```
<aside>
    <div><h3>Aside Title</h3></div>
</aside>
```
- [ ] Headline Links to other articles in an aside should be wrapped in `<article>` tags
```
<aside>
    <article>
        <h4><a>Headline Link to another article</a></h4>
    </article>
</aside>
```
- [ ] Copyright notice in article should be a `<footer>` element rather than a `<p>` [PR](https://github.com/Financial-Times/next-article/pull/1313)

i-like-robots:
what's the status of this?

andygnewman:
Haven't yet had a chance to come back to this following discussion on 'version controlling' ES.

PR #393, opened 2016-05-04
pullquotes as a blockquote wrapped in an aside

andygnewman:
Restructuring article html to have pull quotes as blockquotes within an aside.

As part of this set of issues: https://github.com/Financial-Times/next-es-interface/issues/383


i-like-robots:
Closing because I'm guessing there was no pressing need for this

PR #395, opened 2016-05-04
Exclude podcasts by testing the provenance field

i-like-robots:
Rather than checking tag which may also be (read: _is_) used by CAPI content.


matthew-andrews:
proper job +1

PR #396, opened 2016-05-05
Allow a 1 article difference between the US and EU totals

keirog:
A 1 article delta is pretty common (and expected), and triggers
health alerts all the time. Hopefully this will quieten things down.


keirog:
@i-like-robots @andygnewman

matthew-andrews:
Hmm……  not so sure…… can we just put a timer on this instead…… let it fail for a few interations before raising the alarm instead?

matthew-andrews:
Thought about this some more, I think it's a really important health check. The two regions must be in sync eventually. It's ok for this check to fail (up to say) 10 times in a row but it's not OK for them to have different numbers of articles.  Sorry >.<

andygnewman:
The thought behind having an allowable diff of 1 was that the pollers between regions are likely to be slightly out of sync, so having a diff of 1 is quite probable and leading to noisy alarms.

It would only be if that diff got any larger that it is a sign that one of the clusters isn't behaving.

Your suggestion of an allowable number of contiguous times for it to be out by up to one could be a way to check that it is then getting back in sync OK.

PR #400, opened 2016-05-16
Improve the failover of the out of sync healthcheck

matthew-andrews:
@andygnewman


matthew-andrews:
i think we need a better way of standard debouncers for our health checks :-/

andygnewman:
Nice. Hopefully that should make it less noisy.

PR #402, opened 2016-05-16
If there's no bodyXML coming from V2, delete it so that V1 takes over

matthew-andrews:
probably need a test 😁


matthew-andrews:
or a completely different approach

matthew-andrews:
geronimooooo

PR #403, opened 2016-05-17
Only add bodyXML to article v2 model if it's not false so that deepme…

matthew-andrews:
…rge ignores it if it's falsey


GlynnPhillips:
👍

Issue #397, opened 2016-05-09
Prepare for Video in `articleXML` API change


matthew-andrews:
<img width="722" alt="screen shot 2016-05-09 at 13 21 30" src="https://cloud.githubusercontent.com/assets/825088/15112896/067c9fba-15e9-11e6-9090-7d7ee357925f.png">

```
<a data-asset-type="video" data-embedded="true" href="http://video.ft.com/4877905127001" />
```

To become:-

```
 <ft-content data-embedded="true" url="http://api.ft.com/content/332ecb9e-abb3-31ed-9370-b7202a9b4b97" type="http://www.ft.com/ontology/content/MediaResource" />
```

And the second API request will (initially) return:-

```
{
  "id": "http://www.ft.com/thing/bad50c54-76d9-30e9-8734-b999c708aa4c",
  "type": "http://www.ft.com/ontology/content/Video",
  "webUrl": "http://video.ft.com/4492075574001"
}
```

(In the future this will expand to have more data to make it more useful and potentially even mean we can get rid of our brightcove proxy…… but for now MVP……)

fyi @i-like-robots @andygnewman @TheoLeanse 
(up to us and mr @mmockett to decide between us who does this work 😄 )

andygnewman:
So as MVP we would just need to pre-resolve the url back to the original format(?)

What's the timescale on this?

matthew-andrews:
next few weeks or so

matthew-andrews:
Probably good to start thinking about this :]

andygnewman:
Closed by this https://github.com/Financial-Times/next-es-interface/pull/407

PR #407, opened 2016-05-24
enable video as an ft content element

andygnewman:
@matthew-andrews 
To resolve this forthcoming issue https://github.com/Financial-Times/next-es-interface/issues/397

Just transforms it back to how we expect to see it before the change.
Designed to ensure the change in UPP doesn't break display.


matthew-andrews:
looks good to me!

PR #409, opened 2016-05-26
Pre-resolve ft-content of type MediaResource

andygnewman:
@matthew-andrews 

Could be either video or image, so have catered for both, which led to a rewrite of the image set pre- resolve to abstract out the final step of image transformation in that process.

Plus sorting out all the tests...


matthew-andrews:
looks good, sorry again about this …

PR #413, opened 2016-06-08
Update content-updating.js

adambraimbridge:
Increase the threshold for triggering an alert, because it always fixes itself anyway.

cc @matthew-andrews @laurieboyes


server/checks/content-updating.js
```diff
@@ -37,14 +38,14 @@ module.exports = () => {
...
-					oldest.length = 5;
+				if (oldest.length > threshold) {
+					oldest.length = threshold;
 				}
 				var debug = "\n\nDebug: " + JSON.stringify(unique(oldest));
 				if (unique(oldest).length > 1) {
 					ok = true;
 					checkOutput = "All's well" + debug;
-				} else if (oldest.length > 4) {
+				} else if (oldest.length = threshold) {
```
laurieboyes: The single `=` makes this an assignment, we want to keep the `>`

adambraimbridge: Thanks Laurie. I put these little errors in on purpose to make sure you're paying attention

server/checks/content-updating.js
```diff
@@ -37,14 +38,11 @@ module.exports = () => {
 			.then(function (result) {
 				var uuid = result.hits.hits[0]._id;
 				oldest.unshift(uuid);
-				if (oldest.length > 5) {
-					oldest.length = 5;
```
matthew-andrews: this code needs to be there, otherwise this check won't work :)

matthew-andrews:
I don't think this is needed / achieved what it was intended to.

PR #415, opened 2016-06-10
Add cache purges for updates, deletes

rowanbeentje:
I've had this sitting around for ages, kept meaning to write tests.  I think as we discussed none of the api stuff is currently tested, which I thought was fine – I could set up some basic mocks, etc – but there's not even a setup to mock `fetch()`es on this project so I've given up on that front :)

Tried to follow the rest of the file's es6 approach - yes to template strings and promises, no to arrow functions?


matthew-andrews:
arrow functions are totally allowed now

server/lib/api-v3.js
```diff
@@ -51,11 +55,37 @@ function update(id, opts) {
...
+		})
+		.then(function(res) {
+			if (res.ok) {
+				if (opts.verbose) {
+					logger.info(`event=PURGED uuid=${id}`);
+				}
+			} else {
+				logger.info(`event=PURGE_FAILED uuid=${id}`);
+			}
+			return !!res.ok;
```
matthew-andrews: feel [oddly qualified](https://github.com/whatwg/fetch/issues/18#issuecomment-71345650) to say that `.ok` is a boolean so `!!` is unnecessary 🙃

matthew-andrews:
one stylistic comment but really it's good to go.  merge at will.  💚

rowanbeentje:
I'll leave it for someone else to merge next week as now's probably the wrong time :D

PR #417, opened 2016-06-14
Size the cluster node to r3.xlarge from r3.2xlarge

richard-still-ft:


templates/cluster.json
```diff
@@ -1,5 +1,31 @@
 {
   "AWSTemplateFormatVersion": "2010-09-09",
+  "Parameters": {
+		  "dataNodeInstanceType": {
+        "Type": "String",
```
matthew-andrews: ur indention is rubbish

matthew-andrews:
🚢

PR #422, opened 2016-06-16
Allow syndicated content to come through to Next in a simple form

matthew-andrews:
Don't merge until chatted with @musthemag about legal things.


matthew-andrews:
needs legal

PR #424, opened 2016-06-21
Remove the slideshow-as-an-article hack

matthew-andrews:
- It has been broken for ages.
- It should be fixed properly in V2.


andygnewman:
Nice to tidy up some code. 👍

PR #427, opened 2016-06-23
Add mediaType to the list of ignored taxonomies

i-like-robots:


matthew-andrews:
fails tests, but yes +1 — better to hold off until eu referendum stuff is all finished before deploying though

matthew-andrews:
haha, and now tests

i-like-robots:
Maybe I'll try to fix this locally instead of with the Github editor!

matthew-andrews:
😅

i-like-robots:
proper job

matthew-andrews:
+1 für Montag

PR #428, opened 2016-06-23
Handle body tags with IDs

adgad:
/cc @matthew-andrews @i-like-robots @andygnewman 
Some articles come through with body tags with IDs (eg. https://next-es-interface.ft.com/content/4a105242-2d97-11e6-bf8d-26294ad519fc).

These articles break ads inside articles, since the ad placing logic relies on them being top level elements.


i-like-robots:
LGTM

We should pester the UPP team about this too because that's 💩

adgad:
Pestered UPP and going to merge this now

PR #433, opened 2016-07-01
Add url

### LINKED ISSUE #432: new `url` property …
… which will (initially) be, for normal articles:-

```
/content/<uuid>
```

but for ftalphaville articles will be the original ftalphaville url (because we shouldn't be serving it on Next)

```
http://ftalphaville.ft.com/2016/05/12/2161488/ft-opening-quote-60/
```

eventually this will contain the vanity url for articles.

cc @TheoLeanse @andygnewman

[i think `webUrl` should eventually be deprecated 😁]

matthew-andrews:
Closes #432


server/app.js
```diff
@@ -65,7 +65,8 @@ api.delete('/item', require('./controllers/item'));
 api.post('/proxy', jsonParser, require('./controllers/proxy'));
 
 // Preview what ES interface will do
-api.post('/preview', jsonParser, require('./controllers/api/preview'));
+api.post('/preview', jsonParser, require('./controllers/api/preview').post);
+api.get('/preview/:uuid', require('./controllers/api/preview').get);
```
matthew-andrews: just to make testing one's changes to the model a wee bit easier

andygnewman:
👍

PR #440, opened 2016-07-12
Increase timeout for updating an article 

i-like-robots:
This is a small change to try and reduce the noise of the updater health checks. The continuous updater is often failing when attempting to update a document.

As we are fetching 100 articles at a time we could reindex them as a single request using the `_bulk` endpoint rather than individual requests.


matthew-andrews:
timeout increase seems ok.  `_bulk` sounds interesting.

matthew-andrews:
+1

i-like-robots:
From the docs:

> In the same way that mget allows us to retrieve multiple documents at once, the bulk API allows us to make multiple create, index, update, or delete requests in a single step. This is particularly useful if you need to index a data stream such as log events, which can be queued up and indexed in batches of hundreds or thousands.

The reindex tool I built uses it, we could look at pulling in the code from there.

i-like-robots:
Here's the magic: https://github.com/i-like-robots/elastic-reindex/blob/master/lib/runner.js#L151-L158

PR #442, opened 2016-07-13
Add next-search-read-kmt user

dgem:
We'd like to use ES search as believe we've have followed the instructions correctly ;)


richard-still-ft:
👍 Be gentle lest we revoke your access

PR #443, opened 2016-07-14
Update access policy and add integration tests

richard-still-ft:
Limits writes to a privileged few and should help reduce accidental breaking changes to settings etc.. Adds a development cluster config and some integration tests that run against it.


templates/developmentCluster.json
```diff
@@ -0,0 +1,128 @@
...
+		"dedicatedMasterNodeInstanceCount": {
+			"Type": "Number",
+			"Default" : 3,
+			"AllowedValues" : [2,3,4],
+			"Description" : "Pick a number, any number (between 2 and 4). Normally a 3"
+		},
+		"indexName": {
+			"Type": "String",
+			"Default" : "v3_api_v2-2016-07-01",
+			"AllowedValues" : ["v3_api_v2", "v3_api_v2-2016-07-01"],
```
i-like-robots: I think this should just be `v3_api_v2` - when we bring a new index up it will always be aliased to this name http://next-elastic.ft.com/_aliases

templates/developmentCluster.json
```diff
@@ -0,0 +1,128 @@
...
+							}
+						},
+						{
+							"Effect": "Allow",
+							"Action": "es:*",
+							"Resource": [
+								{ "Fn::Join": [ "", [ "arn:aws:es:", { "Ref": "AWS::Region" }, ":", { "Ref": "AWS::AccountId" }, ":domain/", { "Ref": "AWS::StackName" }, "/", { "Ref" : "indexName" }, "/_search" ] ] },
+								{ "Fn::Join": [ "", [ "arn:aws:es:", { "Ref": "AWS::Region" }, ":", { "Ref": "AWS::AccountId" }, ":domain/", { "Ref": "AWS::StackName" }, "/", { "Ref" : "indexName" }, "/_mget" ] ] },
+								{ "Fn::Join": [ "", [ "arn:aws:es:", { "Ref": "AWS::Region" }, ":", { "Ref": "AWS::AccountId" }, ":domain/", { "Ref": "AWS::StackName" }, "/", { "Ref" : "indexName" }, "/*/_search" ] ] },
+								{ "Fn::Join": [ "", [ "arn:aws:es:", { "Ref": "AWS::Region" }, ":", { "Ref": "AWS::AccountId" }, ":domain/", { "Ref": "AWS::StackName" }, "/", { "Ref" : "indexName" }, "/*/_mget" ] ] }
```
i-like-robots: I think `_count`, `_stats` and `_validate` are legit too - not sure we use any of them in prod (i've used the latter locally a fair bit) but we may in future.

templates/developmentCluster.json
richard-still-ft: Ah OK makes sense.

richard-still-ft:
@i-like-robots I added a specific rule for `_validate` and tests for `validate`, `_count` and `stats` the other latter two are GETs but that should be tightened up soon so worth having them imo

PR #444, opened 2016-07-21
Initial implementation of a bulk API method

i-like-robots:
As the continuous updater batches 100 documents at a time we can overstretch our clusters with the number of requests or reindexes we're asking it to handle simultaneously. The docs say:

> The bulk API makes it possible to perform many index/delete operations in a single API call. This can greatly increase the indexing speed.

Which sounds good! We already use this API when creating a fresh index (with the [reindex tool](https://github.com/i-like-robots/elastic-reindex)).


i-like-robots:
Closed in favour of #561

PR #447, opened 2016-07-22
Add asciifolding to our suggestion filter to cater for diacritics

i-like-robots:
https://www.elastic.co/guide/en/elasticsearch/guide/current/asciifolding-token-filter.html


matthew-andrews:
lóökş ĝō̂od

matthew-andrews:
can we merge this?

i-like-robots:
Yes, but it requires a new index =[

matthew-andrews:
that's ok.  in time.

PR #448, opened 2016-08-02
Initial implementation of video model

i-like-robots:
If a CAPI v2 response is a video then this will intercept and use a different model that will fetch extra data from Brightcove.


server/models/article-v3.js
```diff
@@ -68,6 +68,11 @@ function cleanAndSortStrings (items) {
 		});
 }
 
+function isVideo (articleV2) {
+	return articleV2.type === 'http://www.ft.com/ontology/content/MediaResource'
```
matthew-andrews: worth pulling this a layer above this `article-v3.js` file — perhaps a `negotiation.js` file?

server/models/video.js
```diff
@@ -0,0 +1,71 @@
...
+		}
+
+		const renditions = brightcove.renditions.filter(rendition => {
+			return rendition.videoContainer === 'MP4';
+		});
+
+		return deepmerge(normalizeV2(capiV2), {
+			title: brightcove.name,
+			titles: [],
+			bodyXML: `<p>${brightcove.longDescription}</p>`,
```
matthew-andrews: i would suggest sticking a todo here as `bodyXML`'s are coming _very_ soon to videos in CAPI v2.

server/models/video.js
```diff
@@ -0,0 +1,71 @@
...
+		'command=find_video_by_id',
+		'video_fields=id,name,renditions,longDescription,shortDescription,videoStillURL,customFields',
+		'token=' + process.env.BRIGHTCOVE_APIKEY,
+		'video_id=' + videoId
+	];
+
+	return fetch(`http://api.brightcove.com/services/library?${params.join('&')}`)
+		.then(response => {
+			return response.ok ? response.json() : response.text().then(text => {
+				return Promise.reject({ error: text, status: response.status });
```
matthew-andrews: clever…… but not very readable…… don't you think…… ?

server/models/article-v3.js
i-like-robots: Can do - though if UPP team can be convinced to do this for us then such a thing wouldn't be necessary.

matthew-andrews:
couple of stylistic things and a request to add one `// todo`.  otherwise +1

server/models/video.js
```diff
@@ -0,0 +1,76 @@
...
+			if (response.ok) {
+				return response.json();
+			}
+
+			return response.text().then(text => {
+				return Promise.reject({ error: text, status: response.status });
+			});
+		})
+		.catch(error => {
+			logger.warn({ event: 'BRIGHTCOVE_FETCH_FAIL', region: process.env.REGION, index: 'v3', uuid: videoId, status: error.status });
```
ironsidevsquincy: not logging `error.error`?

server/models/article-v3.js
i-like-robots: I confess, the more I look into this the more worried I am of getting stuck down a warren of refactoring so I've backed away for now =/

matthew-andrews: it's ok, it's not necessary :-)

server/models/video.js
```diff
@@ -0,0 +1,98 @@
...
+			};
+		}),
+		mainImage: {
+			title: `Video still for ${brightcove.name}`,
+			description: '',
+			url: brightcove.videoStillURL,
+			width: 640,
+			height: 360
+		},
+		_lastUpdatedDateTime: new Date().toISOString()
```
matthew-andrews: ah one final thing, I've been adding `url` property to content in advance of these sorts of things getting vanities.  Could you do the same for video?

https://github.com/Financial-Times/next-es-interface/blob/master/server/models/article/normalize-v2.js#L40

matthew-andrews: [ It just needs to be (for now) `/content/{uuid}` and I will add the magic for the vanities later ]

i-like-robots: https://github.com/Financial-Times/next-es-interface/pull/448/commits/6922838866b2daf39cf1ab0bebf25dd5d51a64fc

PR #450, opened 2016-08-04
feat: Set the ‘url’ property to the vanity if one is set up

matthew-andrews:


test/server/models/article-v3-spec.js
```diff
@@ -6,8 +6,15 @@ require('should');
...
-const instance = require('../../../server/models/article-v3');
+const proxyquire = require('proxyquire');
+
+const instance = proxyquire('../../../server/models/article-v3', {
+	'./article/normalize-url': proxyquire('../../../server/models/article/normalize-url', {
+		'@financial-times/n-url-management-api-read-client': {
+			get: () => Promise.resolve({ code: '100' })
+		}
+	})
+});
```
i-like-robots: https://github.com/Financial-Times/next-es-interface/blob/f5c4d5fbb19c2df06e3b6f7e6b00acffa35a0e57/test/server/models/article-v3-spec.js#L31

PR #453, opened 2016-08-05
Remove images when there is no src file

GlynnPhillips:
Rusty xslt skills at play here... 😟 

/cc @andygnewman


andygnewman:
👍  Be worth reingesting that problematic article once this is merged in just to do a final test, but looks good.

PR #457, opened 2016-08-10
Add validation for brightcove response

i-like-robots:
Brightcove does not always return different HTTP statuses.


matthew-andrews:
+1

PR #462, opened 2016-08-11
Add brightcove to provenance field when content has an embedded video

i-like-robots:
This adds the brightcove API endpoint to the provenance field when we manage to resolve an embedded video resource.


matthew-andrews:
+1

test/server/models/article-v3-spec.js
```diff
@@ -70,41 +69,6 @@ describe('Article v3 model', function () {
...
-				.reply(200, dodgyBylineFixtureV2);
-
-			return instance(ARTICLE_ID)
-				.then(response => {
-					response.should.not.have.property('byline');
-				});
-		});
-
-	});
-
```
i-like-robots: handled by other specs

PR #467, opened 2016-08-19
fix: Allow >> in link text for inline links

matthew-andrews:
cc @keirog


keirog:
👍

matthew-andrews:
this is the offending line: https://github.com/Financial-Times/next-es-interface/blob/master/server/models/article/preresolve-links.js#L5

keirog:
@matthew-andrews ^ Better late than never, I hope

matthew-andrews:
yay cool

matthew-andrews:
make sure we update neil, mus and mark etc !

andygnewman:
I see you've bumped up cheerio 0.22. In other apps we'd not moved from 0.19 to 0.20 because of memory issues it was causing. https://github.com/cheeriojs/cheerio/issues/856

The issue hasn't been closed so not clear if 0.22 resolves the issue or not...

keirog:
Oh thanks @andygnewman, I've knocked it down to 0.19

PR #469, opened 2016-08-19
fix: check brightcove response has customfields

i-like-robots:
Fixes

```
TypeError: Cannot read property 'byline' of null
```

=[


i-like-robots:
physical +1 gained

PR #471, opened 2016-08-19
add more response headers to understand better Mashery's 500s

bjfletcher:
Sarah's practical suggestion :)


matthew-andrews:
oh very nice +1

PR #476, opened 2016-08-23
Do not explode when concordances API returns no results

i-like-robots:
When there are no concordances the API returns `{}` rather than `{ "concordances": [] }` so currently things break.


matthew-andrews:
+1

i-like-robots:
Strong gif

matthew-andrews:
<3 concorde

PR #481, opened 2016-08-26
Revert "Merge pull request #478"

i-like-robots:
Because there are plenty of videos in CAPI that lack descriptions etc.

**We can revert this revert when videos have been reingested to CAPI**


matthew-andrews:
yawww i think it's becasue thye need to reingest

matthew-andrews:
so we should revert revert when they do that

i-like-robots:
Yes indeed

matthew-andrews:
+1

matthew-andrews:
go go go go gogo

PR #482, opened 2016-08-30
Ingesting wires / syndicated content

andygnewman:
To Do
- [x] Change api request header to allow policy for wires content
- [x] Add `originatingParty` property to article, defaulted to 'FT', but picking up 'Reuters' when wires content
- [x] When `originatingParty` is not FT, create a mainImage property out of the v1 images array
- [x] In article app, redirect requests for articles with `originatingParty` not equal to 'FT' to classic FT (as an interim until agreed display is finalised in Next). See this [PR](https://github.com/Financial-Times/next-article/pull/1539)


matthew-andrews:
Can we update the schema too?
https://github.com/Financial-Times/next-es-interface/blob/master/schema/content.json

Otherwise +1

matthew-andrews:
wires

PR #483, opened 2016-09-05
feat: Grant access to ES from nextsitemap's application roles (in dev and test) and its deploy user for local dev (dev only)

matthew-andrews:
does this look ok to you?  @TheoLeanse


matthew-andrews:
also cc @richard-still-ft to look at my refactor

PR #487, opened 2016-09-07
add new catch for es get latest

andygnewman:
cc: @matthew-andrews @i-like-robots 

From trying to diagnose the health check alerts this morning we discovered that there was no catch at this point which made it harder to diagnose.


matthew-andrews:
are you sure?

the healthcheck seems to rely on the error:-
https://github.com/Financial-Times/next-es-interface/blob/6f1caced73581570614c75846a0a7fcf39c26218/server/checks/latest-news.js#L36

andygnewman:
Whilst the healthcheck relies on the error there are two different fetches (API search and ES) that could bubble up into the health check error. Knowing which was causing the issue would have helped speedier diagnosis - I focused initially, incorrectly on the API.

Agree this may not be the best approach, but grateful for suggestions.

matthew-andrews:
maybe we can handle the error in the API search?

matthew-andrews:
or change the Health check to expect an empty response rather than a reject in this case?

andygnewman:
So took out that suggested catch and just improved the panic guide description.

Also improved the panic guide for content updating health check.

matthew-andrews:
+1

PR #488, opened 2016-09-12
Log changes in content to Kinesis

TheoLeanse:
/cc @matthew-andrews @richard-still-ft


server/lib/log-changed-content.js
```diff
@@ -0,0 +1,10 @@
+const AWS = require('aws-sdk');
+const kinesis = new AWS.Kinesis({ region: 'eu-west-1' });
+
+module.exports = function logChangedContent (data) {
+	kinesis.putRecord({
```
matthew-andrews: do we want to have any error handling?

server/lib/log-changed-content.js
```diff
@@ -0,0 +1,10 @@
...
+		StreamName: 'nextcontentChangelog',
```
matthew-andrews: picky but elsewhere this is called… `nextContent…`

server/lib/log-changed-content.js
```diff
@@ -0,0 +1,10 @@
...
+		PartitionKey: Math.random().toString(),
```
matthew-andrews: might be sensible to seed the randomness with the `REGION` environment variable and process name?

server/lib/log-changed-content.js
TheoLeanse: what would we do on an error? Log it?

server/lib/log-changed-content.js
TheoLeanse: would need to use a module like this: https://github.com/davidbau/seedrandom ?

server/lib/log-changed-content.js
```diff
@@ -1,12 +1,18 @@
...
 const kinesis = new AWS.Kinesis({ region: 'eu-west-1' });
-const random = require('seedrandom');
+const putRecord = denodeify(kinesis.putRecord);
 
 module.exports = function logChangedContent (data) {
 	const seed = process.env.REGION + process.title || '0';
-	kinesis.putRecord({
+	putRecord({
 		StreamName: 'nextContentChangelog',
 		PartitionKey: random(seed).toString(),
```
matthew-andrews: think this will also need the time otherwise will be same every time…?

server/lib/log-changed-content.js
```diff
@@ -0,0 +1,18 @@
+const logger = require('@financial-times/n-logger').default;
+const random = require('seedrandom');
+const AWS = require('aws-sdk');
+const denodeify = require('denodeify');
+const kinesis = new AWS.Kinesis({ region: 'eu-west-1' });
+const putRecord = denodeify(kinesis.putRecord);
+
+module.exports = function logChangedContent (data) {
+	const seed = process.env.REGION + process.title + new Date().getTime() || '0';
```
matthew-andrews: sorry to be really annoying but this time is in seconds……  might want more granular than that……

also what is `process.title`?  is that the name of the Heroku process or the name of the UNIX process?

TheoLeanse: the unix process! how would I access the heroku process name? can't see it in the config vars

matthew-andrews: `env.DYNO`
https://github.com/Financial-Times/next-metrics/blob/3894d874769b8fe59fd9ae344602b4b5e42545c9/lib/metrics.js#L52

templates/cluster.json
```diff
@@ -139,6 +142,14 @@
         "BucketName": { "Fn::Join": [ "", [ { "Ref": "AWS::StackName" }, "-", { "Ref": "AWS::Region" }, "-backups" ] ] }
       },
       "DependsOn": "nextContentElasticsearchCluster"
+    },
+    "nextContentUpdateDeleteKinesisStream": {
+      "Condition": "IsEU",
+      "Type": "AWS::Kinesis::Stream",
+      "Properties": {
+        "Name": "nextContentChangelog",
+        "ShardCount": 1
```
TheoLeanse: is one enough?

server/lib/log-changed-content.js
```diff
@@ -0,0 +1,18 @@
+const logger = require('@financial-times/n-logger').default;
+const random = require('seedrandom');
+const AWS = require('aws-sdk');
+const denodeify = require('denodeify');
+const kinesis = new AWS.Kinesis({ region: 'eu-west-1' });
+const putRecord = denodeify(kinesis.putRecord);
+
+module.exports = function logChangedContent (data) {
+	const seed = process.env.REGION + process.env.DYNO + Date.now(); // unique by region, dyno and time in ms
```
TheoLeanse: we don't need to test for these being defined and being strings, do we?

TheoLeanse:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;WXSVcRAG&#x2F;2-autoupdating-sitemaps"> Autoupdating sitemaps</a>

templates/cluster.json
matthew-andrews: can we try it and increase if we need?

server/lib/log-changed-content.js
```diff
@@ -0,0 +1,19 @@
...
+const random = require('seedrandom');
+const AWS = require('aws-sdk');
+const denodeify = require('denodeify');
+const kinesis = new AWS.Kinesis({ region: 'eu-west-1' });
+const putRecord = denodeify(kinesis.putRecord);
+
+module.exports = function logChangedContent (data) {
+	if (process.env.REGION !== 'EU') return Promise.resolve(); // only EU needs to log events
+	const seed = process.env.REGION + process.env.DYNO + Date.now(); // unique by region, dyno and time in ms
+	putRecord({
```
matthew-andrews: return

templates/cluster.json
TheoLeanse: yup

server/lib/log-changed-content.js
TheoLeanse: 😢

matthew-andrews:
giphy

PR #489, opened 2016-09-13
fix: return the right response

TheoLeanse:
/cc @matthew-andrews


matthew-andrews:
oops how did we both not spot that!  +1

PR #495, opened 2016-09-15
Tests for changelog logger

TheoLeanse:
/cc @matthew-andrews


test/server/lib/log-changed-content-spec.js
```diff
@@ -0,0 +1,57 @@
...
+const loggerSpy = sinon.spy();
+const logChanges = proxyquire(
+	'../../../server/lib/log-changed-content',
+	{
+		'denodeify': () => putRecordSpy, // HACK: because we are denodeifying the AWS SDK we can just stub denodeify
+		'@financial-times/n-logger': { default: { info: loggerSpy } }
+	}
+);
+
+describe.only('Content changelogger', () => {
```
matthew-andrews: not only

test/server/lib/log-changed-content-spec.js
```diff
@@ -0,0 +1,57 @@
...
+
+	it('should only put records to kinesis in the EU', () => {
+		process.env.REGION = 'US';
+		return logChanges().then(() => {
+			expect(putRecordSpy).not.called;
+		});
+	});
+
+	it('should pass data to kinesis as a string', () => {
+		process.env.REGION = 'EU';
```
matthew-andrews: probably should put this into a `beforeEach` otherwise if mocha decides to run ‘should only put records to kinesis in the EU’ before the others, tests will break.

test/server/lib/log-changed-content-spec.js
TheoLeanse: 😁

matthew-andrews:
+1

PR #497, opened 2016-09-16
feat: Start bringing standfirsts into ES-interface

matthew-andrews:
…… and deprecate the `title` property.


matthew-andrews:
yes it does.

PR #506, opened 2016-09-21
Optimise changelog

TheoLeanse:
/cc @matthew-andrews 

I think there's quite a lot here... 😁 

...not least a node 4 vs node 6 issue


server/lib/es-request.js
```diff
@@ -0,0 +1,85 @@
...
+module.exports = function esRequest (event, uuid, verbose = false, modelledContent) {
+	const URL = `https://${CLUSTER}/v3_api_v2/item/${uuid}`;
+
+	const CONFIG = {
+		DELETE: {
+			esOpts: { method: `DELETE` },
+			purgeOpts: { purgeHard: true },
+			url: URL
+		},
+		UPDATE: {
```
matthew-andrews: `UPDATE_TIMESTAMP`?

server/lib/es-request.js
```diff
@@ -0,0 +1,85 @@
...
+			esOpts: {
+				method: `POST`,
+				body: JSON.stringify({
+					_lastUpdatedDateTime: (new Date()).toISOString() // TODO: or _source._lastUpdatedDateTime? see http://next-elastic.ft.com/v3_api_v2/item/11150514-7e78-11e6-8e50-8ec15fb462f4
+				})
+			},
+			purgeOpts: { verbose },
+			url: `${URL}/_update`
+		},
+		ADD: {
```
matthew-andrews: `UPSERT`?

server/lib/api-v3.js
```diff
@@ -2,93 +2,46 @@
...
-	opts = opts || {};
+function update(uuid, opts = {}) {
 	const verbose = opts.verbose ? true : false;
+	const url = `https://${CLUSTER}/v3_api_v2/item/${uuid}`;
 
-	const url = `https://${CLUSTER}/v3_api_v2/item/${id}`;
-	return articleV3Model(id)
-		.then(function (data) {
-			const issues = sanityCheck(data);
+	const isItemInES = res => {
```
matthew-andrews: i think maybe `isItemInES` isn't either not quite the right name for this function or we should move `signedFetch(url, { method: `GET` })` into it?

server/lib/api-v3.js
```diff
@@ -2,93 +2,46 @@
...
+		// TODO: is it ok to assume json? oficially speaking we should get a 200 if in ES and 404 if not
+		const clone = res.clone();
```
matthew-andrews: i _think_ this isn't needed

server/lib/api-v3.js
```diff
@@ -2,93 +2,46 @@
...
+function update(uuid, opts = {}) {
 	const verbose = opts.verbose ? true : false;
+	const url = `https://${CLUSTER}/v3_api_v2/item/${uuid}`;
 
-	const url = `https://${CLUSTER}/v3_api_v2/item/${id}`;
-	return articleV3Model(id)
-		.then(function (data) {
-			const issues = sanityCheck(data);
+	const isItemInES = res => {
+		// TODO: is it ok to assume json? oficially speaking we should get a 200 if in ES and 404 if not
```
matthew-andrews: yes it is fine to assume json

server/lib/es-request.js
```diff
@@ -0,0 +1,85 @@
...
+			const promises = [
+				purgeCache(uuid, CONFIG[event].purgeOpts),
+				logChangedContent({ event, uuid, url: CONFIG[event].url }, { verbose })
+			];
+
+			return Promise.all(promises).then(() => res.json());
+		});
+};
+
+function purgeCache(id, opts = {}) {
```
matthew-andrews: this is to do with fastly rather than es so could make the file smaller by moving it out?

server/lib/api-v3.js
TheoLeanse: as suggested we'll call it getItem and return undefined if the item isn't there

server/lib/es-request.js
matthew-andrews: i think we decided to have CREATE + UPDATE

bjfletcher:
@TheoLeanse This got my attention :)

> ...not least a node 4 vs node 6 issue

interested to know what was the issue? My interest is that I've been moving apps to use Node 6 from 4.

TheoLeanse:
@bjfletcher not an issue with node, but an issue with my code assuming the wrong version 😀  
We've updated to use node 6 now, and can't see any issues 💃

TheoLeanse:
> i think we decided to have CREATE + UPDATE
Went with that - look ok?

matthew-andrews:
yup that's good!

TheoLeanse:
huzzzah!

PR #510, opened 2016-09-22
[WIP] Create bodyHTML from bodyXML if no V2

andygnewman:
cc: @i-like-robots @matthew-andrews @bjfletcher 

In our quest to deprecate bodyXML fields, this would be a quick hack to ensure that all content had a bodyHTML field.

By this stage the bodyXML has had all tags except `<p>'s` stripped out (in the normalization process), so should be clean and in an HTML state.


andygnewman:
Surprised this change didn't trip up on a test though...

server/models/article-v3.js
```diff
@@ -233,6 +233,7 @@ function applyModel (articleV1, articleV2) {
 		&& articleV1.item.images.length > 0) {
 		const mainImage = mainImageV1(articleV1.item.images);
 		if (mainImage) article.mainImage = mainImage;
+		article.bodyHTML = article.bodyXML;
```
matthew-andrews: This looks like it only get applied for syndicated content?

andygnewman:
Hmm, don't you love multiline if's.

PR #514, opened 2016-09-23
UPDATE -> UPDATE_TIMESTAMP

TheoLeanse:
@matthew-andrews


matthew-andrews:
thanks, i think that's much clearer to the users of the kinesis stream :)

PR #515, opened 2016-09-30
Multi region images

TheoLeanse:
/cc @richard-still-ft @andygout @i-like-robots


richard-still-ft:
👍

richard-still-ft:
@matthew-andrews are you OK with this?

server/lib/multiregion-images.js
```diff
@@ -0,0 +1,4 @@
+module.exports = function multiRegion (imageUrl) {
+	const UUID = /[a-f0-9]{8}-[a-f0-9]{4}-[a-f0-9]{4}-[a-f0-9]{4}-[a-f0-9]{12}/.exec(imageUrl);
+	return `http://prod-upp-image-read.ft.com/${UUID}`;
```
matthew-andrews: `/regex/.exec()` seems to return an array of results (in this case returning the uuid in an array).  i think possibly there needs to be some sort of

```js
if (UUID && UUID[0]) {
  return `http://prod-upp-image-read.ft.com/${UUID[0]}`;
}
console.warn({ event: 'UNABLE_TO_EXTRACT_UUID_FROM_IMAGE', imageUrl })
return imageUrl;
```

matthew-andrews: hmm, i see, your code actually does work because template strings are a wee bit magic.

```
`${['oh', 'hai']}` // oh,hai
```

i think it's safe to keep it as you have it but please add a comment saying that `UUID` is an array and that the template string will do the equivalent of `join(',')`

PR #518, opened 2016-10-03
hacky tagging of new fastft aticles

wheresrhys:


server/lib/sanity-check.js
```diff
@@ -30,12 +30,20 @@ module.exports = (content) => {
...
 
 	// HACK: temporary hack while triaging fastft publish delays
-	if (content.webUrl && content.webUrl.indexOf('/fastft') > -1) {
+	if (content.webUrl && content.webUrl.indexOf('www.ft.com/fastft') > -1) {
 		const id = content.id || 'invalid-id';
 		if (content.metadata && content.metadata.some(tag => tag.idV1 === 'NTlhNzEyMzMtZjBjZi00Y2U1LTg0ODUtZWVjNmEyYmU1NzQ2-QnJhbmRz')) {
 			logger.warn({event: "FASTFT_CONTENT_MONITOR", id, tagged: true})
 		} else {
 			logger.warn({event: "FASTFT_CONTENT_MONITOR", id, tagged: false})
+			content.metadata.push({
```
matthew-andrews: hmm i know this is a hack but it shouldn't go in the `sanity-check` file…… could we move this to be with other metadata hackery (and add a test) in the metadata-transformer file?

https://github.com/Financial-Times/next-es-interface/blob/master/server/lib/metadata-transformer.js

wheresrhys: That file doesn't have access to webUrl (obviously it could by adding an additional parameter, but then the hack becomes a bit less isolated)

It's nasty having it here too I know, but the fix in wordpress is fairly imminent I believe, so hopefully won't live for very long

matthew-andrews: hmm, ok then, but still think it should have a test and not go into the sanity-check file…… :-/ 

cc @i-like-robots for a second opinion

i-like-robots: I agree - the sanity check should probably stay 'sane'.

It could go in [this module](https://github.com/Financial-Times/next-es-interface/blob/master/server/models/article/metadata-v1.js)

matthew-andrews:
+1 thanks you

PR #524, opened 2016-10-10
Make v1 text only version of bodyXML the default bodyHTML

andygnewman:
cc: @i-like-robots, @matthew-andrews 

This seems like the simplest way to have a bodyHTML whenever the source has a bodyXML.

Sets bodyHTML to the stripped down version of the V1 bodyHTML (as currently used for bodyXML), which will then get overwritten by the more sophisticated V2 version if there is a V2.


i-like-robots:
Perfect. This functionality differs from how we deal with the V2 response (storing XML almost as-is) so I think this was probably the right thing to do all along.

andygnewman:
Thanks, will merge this in.
Will leave the removal of the existing XML fields until;
- I've found and resolved all the dependencies
- Which will in turn need to wait on this cycling through all the content

i-like-robots:
+1

Issue #522, opened 2016-10-06
Content model debt


i-like-robots:
A place to list gripes or commonly repeated patterns that should be stored in Elasticsearch.

A few ideas to begin:
### Stop storing `summaryXML` or `bodyXML`.

Nobody should be using it anymore and wastes bandwidth when not specifying fields.
https://github.com/Financial-Times/next-es-interface/issues/525
### Store primary tag

The logic to pull out the primary tag is used in multiple apps so we should just store it.
https://github.com/Financial-Times/next-es-interface/issues/526
### Store one alternative title

We only ever pull out the first, why store many?
https://github.com/Financial-Times/next-es-interface/issues/527

### ~Stop storing `annotations`~

~For the same reasons as the XML fields, they can always be re-added when they become useful to somebody.
https://github.com/Financial-Times/next-es-interface/pull/528~

### Re-name the index

The origins of `v3_api_v2` are lost. Let's just name the index `content`.

matthew-andrews:
> Store one alternative title

The model in UPP has changed a bit.  I think we should deprecate the `headlines` and `summaries` arrays in favour of `standfirst` and whatever-they-called-the-alt-headline.

andygnewman:
Started work on the XML piece (https://github.com/Financial-Times/next-es-interface/issues/525) - I think AMP will be the biggest hurdle on that one.

Will start going through the others also.

i-like-robots:
Closing this as we have issues for each individual point.

PR #528, opened 2016-10-11
Stop storing annotations

andygnewman:
cc: @i-like-robots 

As per this issue.

https://github.com/Financial-Times/next-es-interface/issues/522


i-like-robots:
Smashing!

i-like-robots:
Ship it?

server/models/article/metadata-v2.js
```diff
@@ -1,18 +0,0 @@
-'use strict';
-
-const SILLY_PREFIX = /http:\/\/(?:www|api)\.ft\.com\/things?\//;
```
matthew-andrews: :)

PR #532, opened 2016-10-13
log difference between versions

wheresrhys:
The jist is that for any property on the model it checks if the JSON representation (or the raw value if a primitive) has changed at all. No crazy traversing of properties at this stage


wheresrhys:
I've changed to output a simpleDiff property

Issue #525, opened 2016-10-11
Deprecate XML fields in favour of HTML properties


andygnewman:
We currently have both partially cleaned up XML properties and transformed HTML properties.

It is proposed that our ES index should only retain the HTML versions.

To achieve this we need to;

## 1. Ensure there is always a bodyHTML
bodyHTML has only been populated when there is a V2 version of the content, with consuming apps having to fallback to the stripped down (text only) V1 bodyXML.
In future all article content should have a bodyHTML, which takes the form of the fallback to the V1 bodyXML when no V2 version exists.
- [ ] bodyHTML consistently present for all article content (https://github.com/Financial-Times/next-es-interface/pull/524)

## 2. Wait until all article content has been reingested through continuous update cycle
- [ ] Approx 19/10/2016

## 3. Remove dependencies on XML fields from consuming apps
- [ ] `next-article` https://github.com/Financial-Times/next-article/pull/1634
- [ ] `n-content-decorator` https://github.com/Financial-Times/n-content-decorator/pull/10
- [ ] `alphaville-es-interface` https://github.com/Financial-Times/alphaville-es-interface/blob/master/lib/bodyHTML.js
- [ ] any others?

## 4. Remove XML fields from ES article model
- [ ] put PR for removal here...

i-like-robots:
Update:

EU index has 2973 docs without a `bodyXML` field... so nearly there

andygnewman:
Did you mean that only 2973 don't have a `bodyHTML` field rather than `bodyXML`

i-like-robots:
Yes

i-like-robots:
Alphaville are going to maintain their own ES instance.

i-like-robots:
I think we can now do this...

PR #533, opened 2016-10-13
Add bodyHTML to video and podcast models

andygnewman:
cc: @i-like-robots 

Only had bodyXML, but as wanting to deprecate XML property in favour of HTML, need to add these.
as per step 1 of https://github.com/Financial-Times/next-es-interface/issues/525

Will remove the XML property in future PR once dependencies have been resolved.


i-like-robots:
Haha, bugger - good catch!

PR #536, opened 2016-10-17
Add: ratio property to main image

andygnewman:
cc: @i-like-robots 

Part of content data model work.

The main image ratio of width / height is used for presentational purposes in cards / teasers and is being calculated in n-content-decorator.

Seems like a good candidate for resolving on ingest.


i-like-robots:
There's no harm in adding it as we know the values - though it's also only a very tiny piece of code to add where it matters. 

Remember to add to the mapping!

PR #540, opened 2016-10-24
Refactor search updater spec

i-like-robots:
This untangles the search updater process spec from the notifications API spec. This also adds new tests for V2 notifications (which are transformed into V1 style) and adds handling for any HTTP errors.

Work can then continue on refactoring these modules as they're some of our very oldest code (no fetch, no promises!).


andygnewman:
👍

PR #541, opened 2016-10-24
Refactor get notifications and search updater

i-like-robots:
Continuing from #540 

- Refactors get notifications module to use fetch and promises
- Refactors notifications handler (nee search updater) to be simpler (courtesy of some ES6 sugar) and provide more consistent logging
- Refactors notifications poll (nee update search) process to be simpler and includes basic spec

Next: re-factor notifications format (and de-dupe, checking splunk we can reingest an item 3 times in a single poll) and then onto the continuous updater.


andygnewman:
Looks very clear, will let you sort out the linting though.

server/processes/search-updater.js
```diff
@@ -0,0 +1,72 @@
+const REGION = process.env.REGION === 'US' ? 'US' : 'EU';
```
i-like-robots: this module should be named `handle-notifications`

server/lib/get-notifications.js
```diff
@@ -1,105 +1,59 @@
...
-		return notifications.links[0].href;
-	} else {
-		return false;
-	}
+	return notifications;
 }
+
+module.exports = function (since, options) {
+	const url = getNotificationsUrl(options);
+	return getNotifications(`${url}?apiKey=${process.env.API_KEY}&feature.blogposts=on&since=${since}`);
```
i-like-robots: assemble the query string in a more elegant way

server/poll-notifications.js
```diff
@@ -0,0 +1,17 @@
+const searchUpdater = require('./processes/search-updater');
+const args = require('minimist')(process.argv.slice(2));
+const UPDATE_INTERVAL = args.updateInterval || process.env.UPDATE_INTERVAL || 60 * 1000;
```
i-like-robots: drop the env var, `parseInt`

server/poll-notifications.js
```diff
@@ -0,0 +1,17 @@
...
+const API_VERSION = parseInt(args.apiVersion) || 1;
```
i-like-robots: missing radix

i-like-robots:
@andygnewman I've set up some splunk queries to keep an eye this... reckon we're good to 🚢 ?

Issue #197, opened 2015-11-10
Deprecate read endpoints


i-like-robots:
The web interface and some of its endpoints are likely reaching the end of their usefulness. Applications [using these endpoints are few](https://github.com/search?utf8=%E2%9C%93&q=next-es-interface.ft.com+user%3AFinancial-Times&type=Code&ref=searchresults) and there is probably little reason they cannot use Elasticsearch or CAPI directly instead of this middleman.

Endpoints with potential to be removed: 

- [ ] /things
- [ ] /content
- [ ] /brand and /sections and /authors
- [ ] /oldest

matthew-andrews:
i think this is a bit ambitious………

PR #544, opened 2016-10-25
Remove unused endpoints

i-like-robots:
- `/proxy` and `/debug` 
- `/oldest`
- `/authors` and `/brands` and `/sections`
- `/content`

Couldn't find any usage of these endpoints.


i-like-robots:
CC #197

PR #547, opened 2016-10-25
Refactor continuous updater into oldest poller 

i-like-robots:
This splits the continuous updater into separate, tested modules and makes logging more consistent with other processes.


andygnewman:
Looks good, pretty simple to follow. 👍

PR #548, opened 2016-10-26
Add FT Tech Tonic Podcast to podcast feeds

mmockett:
New show - gun be lit.


mmockett:
@i-like-robots still waiting on Mus for a stream to map to but presume we can do this bit independently?

i-like-robots:
Hmmm... well, we could ship this but then we'd need to manually reingest it all again later anyway. We really need the tag adding to [the metadata](https://github.com/Financial-Times/n-podcast-mapping).

mmockett:
Alright - will add that first.

i-like-robots:
Requires https://github.com/Financial-Times/n-podcast-mapping/pull/7

mmockett:
@i-like-robots is this good to go now then?

i-like-robots:
I need to manually release a new version of n-podcast-mapping - I'll do so shortly

i-like-robots:
v1.0.5 has been released

PR #549, opened 2016-10-26
Refactor latest news process

i-like-robots:
Refactors the latest news poller:

- separates into smaller modules with basic specs
- only fetches the content from SAPI we can deal with (articles, blogs)
- skips the modelling/sanity check step to avoid duplicating effort (we'll still get the logs if 'bad' content is sent for ingestion)


server/checks/latest-news.js
```diff
@@ -1,62 +1,68 @@
-'use strict';
+const region = process.env.REGION === 'US' ? 'US' : 'EU';
+const interval = 1 * 60 * 1000;
+const errorThreshold = 3;
```
andygnewman: Capitalize constants?

i-like-robots: done!

andygnewman:
👍

PR #557, opened 2016-10-31
Fix: consistency in attributes model

andygnewman:
@i-like-robots 

In first implementation of headshots we put in 
```
attributes: [
    { headshot: "headshotfileref" }
]
```

However on looking at other attributes, eg. company attributes, they are in the format
```
attributes: [
    { key: "wsod_key", value: "us:XOM" }
]
```

This PR amends the author headshot code to bring it in line
```
attributes: [
    { key: "headshot", value: "headshotfileref" }
]
```


test/server/models/article/metadata/author-headshots-spec.js
```diff
@@ -144,9 +146,9 @@ describe('Metadata - Author Headshots', () => {
 				.then(() => {
 					expect(nock.isDone()).to.be.true;
 					expect(metadata[0].attributes.length).to.equal(1);
-					expect(metadata[0].attributes[0].headshot).to.equal('fthead:has-headshot');
+					expect(metadata[0].attributes[0].value).to.equal('fthead:has-headshot');
 					expect(metadata[1].attributes.length).to.equal(1);
-					expect(metadata[1].attributes[0].headshot).to.equal('fthead:wolfgang-munchau');
+					expect(metadata[1].attributes[0].value).to.equal('fthead:wolfgang-munchau');
```
i-like-robots: I reckon these assertions are tested thoroughly in the previous tests

i-like-robots:
Ah, darn - never mind +1

andygnewman:
At least it's better to have spotted it now than when we've started using it.

PR #560, opened 2016-11-01
Ui makeover

gvonkoss:
WIP!


views/partials/home.handlebars
```diff
@@ -0,0 +1,4 @@
+<button class="o-buttons" onclick="location.href='./delete'">Delete Content</button>
+<button class="o-buttons" onclick="location.href='./reingest'">Reingest Content</button>
+<button class="o-buttons" onclick="location.href='./tme'">Reingest Tag</button>
+<button class="o-buttons" onclick="location.href='./date'">Date</button>
```
i-like-robots: Make these links?

views/partials/delete.handlebars
```diff
@@ -0,0 +1,7 @@
+<h2>Delete Content</h2>
+<label for="o-forms-standard" class="o-forms__additional-info">Remember to delete content from <span>both</span> clusters!</label>
```
i-like-robots: This should just be a paragraph

views/partials/delete.handlebars
```diff
@@ -0,0 +1,7 @@
...
+<input type="text" id="uuid" placeholder="The uuid thing to delete" class="o-forms__text"/>
```
i-like-robots: and the contents of the placeholder should be the label =]

views/partials/delete.handlebars
```diff
@@ -0,0 +1,7 @@
...
+<br></br>
+
+<button class="o-buttons deleteItem">Delete</button>
+<button class="o-buttons" onclick="location.href='./home'">Back</button>
```
i-like-robots: linkedy link

views/partials/delete.handlebars
gvonkoss: But the place holder is uuid, isn't it?

i-like-robots: Labels describe the input, a placeholder is an example, E.G.:

```html
<label for="email-field">Email address</label>
<input id="email-field" placeholder="you@example.com">
```

So I think should be:

```html
<p>Remember to delete content from <span>both</span> clusters!</p>
<label for="uuid">UUID of content to delete</label>
<input placeholder="00000000-0000-0000-000000000000">
```

May also be worth validating the format (`pattern="[a-f0-9]{8}-[a-f0-9]{4}-[a-f0-9]{4}-[a-f0-9]{4}-[a-f0-9]{12}"`)

views/index.html
```diff
@@ -0,0 +1,79 @@
+<!DOCTYPE html>
+<html>
+	<head>
+		<meta charset="utf-8">
+		<title>{{title}}</title>
+		<script src="//cdn.polyfill.io/v2/polyfill.min.js?features=fetch"></script>
```
i-like-robots: I wonder if we should cut-the-mustard and display an error if a browser is unsupported, what do you think?

i-like-robots:
Thanks for your hard work on this! Looks really good, just a couple of small comments

views/index.html
gvonkoss: I put in the ctm, just not 100% sure it is right, or how to test it locally. Thoughts?

i-like-robots:
👍

PR #561, opened 2016-11-01
Add update, delete and bulk modules using elasticsearch client

i-like-robots:
Moving towards implementing the bulk method I've taken time to study our ES interactions. The current modules are OK but tricky to extend.

I've reproduced the current functionality of `the es-request` and `api-v3` modules, splitting into simpler, task-based modules using the official client which simplifies interacting with ES.

This also implements both throttle and bulk on the oldest poller (nee continuous updater) using the new modules!


andygnewman:
Looks very nice, clean and easy to read 👍

PR #565, opened 2016-11-02
deleted many things

gvonkoss:


i-like-robots:
Small comments but otherwise +1000

i-like-robots:
🚢

PR #569, opened 2016-11-03
S3 lifecycle policy

sjparkinson:
Add a lifecycle policy to the Elasticsearch snapshot S3 bucket.

Should help to keep the bucket clean.

Delete objects after 7 days of creation.

https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-s3-bucket-lifecycleconfig-rule.html


i-like-robots:
CC @richard-still-ft @TheoLeanse

TheoLeanse:
👍  coool

sjparkinson:
Mind if I add template validation to the Makefile in another PR?

matthew-andrews:
https://github.com/Financial-Times/next-es-interface/pull/569#issuecomment-258162316

pretty sure @TheoLeanse set that up in another project  — maybe time to extract and put into n-makefile?

sjparkinson:
The convention of where people put the templates however could be an issue.

TheoLeanse:
a validation attempt elsewhere for what it's worth: github.com/Financial-Times/next-sitemaps/blob/master/Makefile#L10-L11
template directory and name is just hard coded there

PR #571, opened 2016-11-03
Automate out of sync ES clusters

leggsimon:
@i-like-robots


leggsimon:
Closing because mess 😞

PR #572, opened 2016-11-03
Combine V1 and V2 notifications pollers and de-dupe notifications

i-like-robots:
This joins the previously separate CAPI V1 and V2 notifications pollers into a single stream and deduplicates the notifications received. 

The motivation being that when investigating publishing errors (by searching by UUID in Splunk) it's clear we model and ingest content items multiple times per poll. The most I've seen is 6 times!


andygnewman:
LGTM, neat implementation of filter to get the uniques. 👍

PR #586, opened 2016-11-08
Refactor IDs API controller and add a spec

i-like-robots:
I've also introduced query param validation which I will extract in a following PR. This is so we can display more helpful messaging in the help UI =]

screen shot 2016-11-08 at 12 39 32


andygnewman:
Looks good to me. 👍

PR #587, opened 2016-11-08
use new image service

JakeChampion:
We launched the new image service 6 days ago, I'm currently going around FT projects and helping them migrate to it.

The new Image Service includes a lot of improvements:
- The application is multi-region with failover
- We use stale-on-error and stale-while-revalidate cache extensions to ensure users get stale content if the application errors or has to revalidate an image
- We serve much smaller WebP and JPEG XR images to supported browsers
- We’ve seen a reduction in file size for a large number of image requests
- We’re now serving images over H2


i-like-robots:
Thanks @JakeChampion ! Looks like the test for the get image size module is calling through to the real API - happy to pick this up if you're busy.

JakeChampion:
@i-like-robots That would be very kind of you, thanks! :-)

i-like-robots:
No worries

i-like-robots:
Updated

PR #593, opened 2016-11-11
Convert the development Elasticsearch cluster template to YAML

sjparkinson:
Took the time to clean up a template and get an idea of the cluster setup.

The resource and parameter name changes do mean existing stacks can't really be updated with this, I believe, so I've left the original for now until we can create new stacks from existing snapshots.


sjparkinson:
Tested a couple of times in infra-prod, tried in infra-dev but it looks like the policy means this can only really be used in the account with all the IAM users and roles unless I'm mistaken.

i-like-robots:
The YAML is definitely an improvement... but I'm afraid I don't completely understand it... @richard-still-ft ?

sjparkinson:
Closing this, will open another when the right time comes (spinning up a new cluster to upgrade to v2 comes to mind).

PR #603, opened 2016-11-22
Add CPU CloudWatch alarm to the ES cluster

sjparkinson:
Add a CloudWatch alarm to keep an eye on the clusters CPU usage.

Pending stack updates:

image

(Bucket change is from #569, not deployed yet.)


sjparkinson:
@richard-still-ft @i-like-robots I've now updated both stacks.

The buckets have a 7 day lifecycle policy, and the stacks have a CPU alarm pointing at the next-search-alarms SNS topic (that has a pagerduty subscription).

PR #605, opened 2016-11-23
Add Next-style url to Video model

ironsidevsquincy:
@andygnewman


andygnewman:
Noted that the next style url currently gets redirected back to video.ft.com.
👍  looking forward to seeing your work on a video article page

ironsidevsquincy:
Yep, that's the next stage, remove the redirect and have a basic video 'article'

PR #607, opened 2016-11-25
Use eager-fetch to get brightcove data

ironsidevsquincy:
Retry 3 times

Also,  add duration to video attachment, and story package

@i-like-robots


server/models/video.js
```diff
@@ -72,7 +70,8 @@ function applyModel (brightcove) {
 				mediaType: 'video/mp4',
 				url: rendition.url,
 				width: rendition.frameWidth,
-				height: rendition.frameHeight
+				height: rendition.frameHeight,
+				duration: rendition.videoDuration
```
ironsidevsquincy: Need the video duration so adding this in too, does that gel with the data model @i-like-robots?

i-like-robots: Nope, but I don't think that matters - we don't use it for search or anything where mapping optimizations are required.

server/models/video.js
```diff
@@ -52,18 +53,15 @@ function applyModel (brightcove) {
 	);
 
 	return {
-		// TODO: this should come from CAPI soon
-		title: brightcove.name,
+		// TODO: these should come from CAPIv2
 		alternativeTitles: {},
-		// TODO: this should come from CAPI soon
 		bodyXML: `<p>${brightcove.longDescription}</p>`,
 		bodyHTML: `<p>${brightcove.longDescription}</p>`,
```
i-like-robots: Could we take the `bodyXML` field from CAPI?

ironsidevsquincy: I did that, but then noticed that its format is different to the format of 'normal' articles, e.g. it doesn't have a wrapping `<p>`

server/models/video.js
ironsidevsquincy: ahh, but I'd like to use it for search, e.g. get all video content that is less than a minute long. Maybe it should live at the top level, `videoDuration` or something?

server/models/video.js
i-like-robots: Could regex? Whatever you think is better anyway

PR #614, opened 2016-11-29
Check video custom fields are not null or undefined before transforming

i-like-robots:
`brightcove.customFields=null` which trumps the default value used in destructuring assignment


andygnewman:
👍

PR #620, opened 2016-11-30
Extract main image

i-like-robots:
This is really some prep work for the refactored model format that's on its way.

cc @andygnewman @leggsimon


server/models/article/main-image-v1.js
```diff
@@ -0,0 +1,15 @@
+module.exports = function (articleV1) {
+	const image = Array.isArray(articleV1.item.images)
```
leggsimon: would `item` here ever be `undefined` resulting in an error?

i-like-robots: No - this process would explode long before this point if CAPI were to provide some malformed content!

andygnewman:
Nice

PR #622, opened 2016-12-01
Combine Sarah Gordon Author and Brand

mmockett:
Ensure that Sarah Gordon's Author page displays instead of Brand.


mmockett:
@andygout - look good to you?

PR #624, opened 2016-12-02
added more info to data sent to update stream

wheresrhys:
@i-like-robots


server/lib/elastic/delete.js
```diff
@@ -1,11 +1,24 @@
...
+
+module.exports = function (uuid) {
+	return get(uuid)
+		.then(oldContent => {
+			return client.delete({
+				id: uuid,
+				type: 'item',
+				index: 'v3_api_v2'
+			})
+				.then(() => logChangedContent({ uuid, changes: {oldContent}, event: 'DELETE' }));
```
i-like-robots: Do we need to log the changes for update/delete? No harm in having it, just unsure of the use case?

wheresrhys: yep, so consumers can identify streams which contained the article

Issue #553, opened 2016-10-28
Refactor article model


i-like-robots:
It's gotten a little out of hand... 

We normalize the responses and merge them... then pluck some V1 properties instead. That's not so bad but the async work is tricky... sometimes it's conditional, sometimes it isn't. Perhaps we can do something a little cleaner with some sugar:

``` js
//
// model.js
//
function setProperty (property, value) {
    if (value instanceof Promise) {
        this.async.push(value);

        value.then((result) => {
            this.model[property] = result;
        });
    } else {
        this.model[property] = value;
    }
}

class Model {
    constructor (schema) {
        this.schema = schema;
        this.model = {};
        this.async = [];
    }

    apply (property, value) {
        if (this.schema.hasOwnProperty(property)) {
            setProperty.call(this, property, value);
        } else {
            throw new Error(`No property named "${property}" in schema`);
        }
    }

    lock () {
        return Promise.all(this.async).then(() => Object.freeze(this.model));
    }
}

module.exports = Model;

//
// article.js
//
const schema = require('../schema/content.json').mappings.item;
const Model = require('./model');

const article = new Model(schema);

// .. go fetch data...

// simple values...
article.apply('publishedDate', normalizedV1.publishedDate);

// async values...
article.apply('metadata', metadataV1(normalizedV1.metadata));

return article.lock().then(es.update);
```

This should be rolled out to the video and podcast models also so that it easier to context switch between them.

i-like-robots:
Work on this has begun! Check progress in the matth/async-article branch.

- [x] extract main image code #620 
- [ ] extract metadata async code
- [ ] extract XML processing async code

PR #628, opened 2016-12-06
Reorganise <ft-content> resolution modules 

i-like-robots:
This refactors the `<ft-content>` tag resolution modules out of the article model. Clearing the way for #553.

The PR is inflated as I've reorganised the modules too so most of the changes are updated dependency paths.

If the tag transforms or xslt steps throw an error then the modelling will fail immediately. Looking in Splunk yields 0 failures of this type within the last 7 days.

[Sanity check](https://github.com/Financial-Times/next-es-interface/wiki/Sanity-checking-model-changes) comparing 45 documents is 🆗


i-like-robots:
CC @leggsimon @andygnewman

andygnewman:
Looking good 👍

PR #629, opened 2016-12-09
A few new mapping for some tag deletions / renames

laurieboyes:
@i-like-robots don't know how much interaction you have with this bit of es-interface but @ ing you just in case you're interested.

I think the item API has other uses, but as far as I'm concerned, this change is so that the nightly myFT sync job will update it so that people who previously followed the deleted topic will instead follow the new, mapped one.


i-like-robots:
FYI this will take ~5 days to update all of our existing indexed content.

👍

PR #633, opened 2016-12-12
Log changed fields to splunk

i-like-robots:
Currently we're logging a pretty useless `[object Object]` so this pulls out the changed, top-level fields when a diff is provided, E.G.

```
event=CONTENT_UPDATE uuid=abc-123 changes="title,publishedDate,metadata"
```

CC @commuterjoy @andygnewman


server/lib/log-changed-content.js
```diff
@@ -7,7 +7,17 @@ const putRecord = denodeify(kinesis.putRecord.bind(kinesis));
...
+
+		// only log changed fields to splunk
+		if (data.changed && Array.isArray(data.changed.diff)) {
+			changes = data.changed.diff
+				.map((change) => change.path ? change.path[0] : null)
+				.filter((item, i, mapped) => mapped && mapped.indexOf(item) === i)
+				.join();
+		}
+
+		logger.info(`event=CONTENT_${data.event} uuid=${data.uuid} changes="${changes}" region=${process.env.REGION}`);
```
ironsidevsquincy: these can be an object of key/values (then will also take care of handling splunk-unfriendly characters)

i-like-robots: Ah, I wondered if there was a difference - I've always favored the string but without knowing why... very useful to know!

PR #641, opened 2016-12-13
Default log args

i-like-robots:
Also bumps the NHT beta version because https://github.com/Financial-Times/n-heroku-tools/issues/436


server/models/article/ft-content/media-resources.js
```diff
@@ -43,7 +41,7 @@ module.exports = (xml) => {
 				$tag.replaceWith(replacement);
 			})
 			.catch((err) => {
-				logger.warn(`event=PRERESOLVE_MEDIARESOURCE_FAILED error=${err.toString()} region=${REGION}`);
+				logger.warn(`event=PRERESOLVE_MEDIARESOURCE_FAILED error=${err.toString()}`);
```
ironsidevsquincy: fyi, n-logger is clever enough to pull out useful error data if you pass it the error object directly, e.g.

```logger.warn({ event: 'PRERESOLVE_MEDIARESOURCE_FAILED' }, err });```

PR #665, opened 2017-01-06
Add primary section fallback

ironsidevsquincy:
i.e. if no primary section is set, set the 'first' section as primary

_Should_ only effect video content, as there is currently no way of setting 'primariness' to tags, whereas all normal methode content _should_ have a primary section tagged (couple of _shoulds_ in there, so might have unexpected consequences...)

@andygnewman


andygnewman:
I would suggest only doing the primary Section fall back if there are no other primaries set as on an article it may be deliberate that a primary Section has not been set, but other primaries may be set.

andygnewman:
If I was being picky I'd suggest a test where there is a primary theme, but no primary section to ensure that it's not being set in that circumstance.

ironsidevsquincy:
Added

andygnewman:
👍

PR #666, opened 2017-01-06
Retain: Company links whitespace (where reqd)

andygout:
cc @i-like-robots @laurieboyes


server/models/article/ft-content/company-links.js
```diff
@@ -14,13 +14,15 @@ module.exports = (xml) => {
 		const $link = $links.eq(i);
 		const companyUuid = $link.attr('url').split('/').pop();
 		const companyText = $link.text().trim();
+		const companyTextLeadingWhitespace = $link.text().match(/^\s.*/) ? ' ' : '';
+		const companyTextTrailingWhitespace = $link.text().match(/.*\s$/) ? ' ' : '';
```
laurieboyes: (Pedantry warning - feel free to ignore)

If you're interested in shaving bytes off your regex, I don't think you need the `.*` bit, as the \s will still match.

E.g. the first one could be `.match(/^\s/)`

Issue #669, opened 2017-01-06
v3_api_v2 -> content


i-like-robots:
Our content index is now available with the alias `content` as the origins of the `v3_api_v2` name have been long forgotten.

https://github.com/search?q=org%3AFinancial-Times+v3_api_v2&type=Code

PR #670, opened 2017-01-06
Re-name all references to 'v3_api_v2' to 'content'

i-like-robots:
As per #669


i-like-robots:
CC @sjparkinson (not sure the changes made to CF template actually matter)

sjparkinson:
Allowed values _might_ have an issue with the existing parameter values, but that's about it.

i-like-robots:
I found out this _does_ have implications for permissions as most roles are tied to the `v3_api_v2` index name.

PR #673, opened 2017-01-09
Auto remove invalid

i-like-robots:
After the change in #671 to strip empty paragraphs from V1 responses we are finding lots of useless content within our indices thus triggering lots of warnings from the `content-updating` healthcheck.

This attempts to make the check self-healing (like out of sync) by testing whether the item it is stuck on is invalid and if so, will attempt to delete it.


i-like-robots:
CC @leggsimon

PR #674, opened 2017-01-09
Remove invalid items from both clusters 

i-like-robots:
The invalid item removal introduced in #673 is working well but because the content is only removed from the cluster the app is tending it's causing out of sync errors!

This changes it to delete the item from _both_ clusters.


i-like-robots:
CC @leggsimon @lc512k

PR #675, opened 2017-01-10
Get decorated data for video content from the new video service

ironsidevsquincy:
@i-like-robots


server/models/video.js
```diff
@@ -141,10 +156,11 @@ module.exports = function videoModel (capiV2) {
 
 	const thingsToDo = [
 		fetchConcordances(capiV2.annotations),
-		fetchBrightcove(videoId)
+		fetchBrightcove(videoId),
+		fetchVideo(videoId)
 	];
 
-	return Promise.all(thingsToDo).then(([ concordances, brightcove ]) => {
+	return Promise.all(thingsToDo).then(([ concordances, brightcove, { tags = [] } = {} ]) => {
```
i-like-robots: I'll hold off adding code coverage! Just as an FYI the parser used by Istanbul can't cope with this syntax still

i-like-robots:
LGTM 🚢

PR #680, opened 2017-01-12
Add publish reference field

sjparkinson:


sjparkinson:
Question about that `content.json` schema, how's that get loaded into Elasticsearch?

sjparkinson:
To answer my own question, https://www.elastic.co/blog/found-elasticsearch-mapping-introduction#how-to-provide-a-mapping.

So provided when you create the index. To update a request like the following should do it:

```
PUT /content/_mapping/item

{
  "properties": {
    "publishReference": {
      "type": "string",
      "index": "not_analyzed"
    }
  }
}
```

i-like-robots:
Perfect, thanks!

Unfortunately it's not possible to update an existing index (you must close then reopen to do so but these commands are not available on AWS) but for this simple case that's not a problem.

PR #681, opened 2017-01-13
Add publish reference to the changed content log

sjparkinson:
Should result in Splunk logs like:

```
event=CONTENT_UPDATE uuid=0eecd666-d96a-11e6-944b-e7eb37a6aa8e changes='bodyXML,bodyHTML,publishReference,annotations' publish_reference=tid_dcmi1r6wq1 region=US appId=1801840e-5f32-4c86-86e3-6a6ece40f7bb appName=ft-next-es-interface-us level=info
```


server/lib/log-changed-content.js
```diff
@@ -17,7 +17,14 @@ module.exports = function logChangedContent (data = {}, opts) {
...
 
-		logger.info(`event=CONTENT_${data.event} uuid=${data.uuid} changes="${changes.join()}"`);
+		let logMessage = `event=CONTENT_${data.event} uuid=${data.uuid} changes="${changes.join()}"`;
+
+		// Log publish reference if available.
+		if (data.changes && data.changes.newContent && data.changes.newContent.publishReference) {
+			logMessage = logMessage + `publish_reference=${data.changes.newContent.publishReference}`;
+		}
+
+		logger.info(logMessage);
```
ironsidevsquincy: n-logger can take an object and it will turn them into splunk fields (and sanitise them), e.g.

```
logger.info({ event: `CONTENT_${data.event}`, uuid: data.uuid, ...});
```

server/lib/log-changed-content.js
```diff
@@ -17,7 +17,20 @@ module.exports = function logChangedContent (data = {}, opts) {
...
-		logger.info(`event=CONTENT_${data.event} uuid=${data.uuid} changes="${changes.join()}"`);
+		let message = {
+			event: 'CONTENT_' + data.event,
+			uuid: data.uuid,
+			changes: changes.join()
+		};
+
+		// Log publish reference if available.
+		if (data.changes && data.changes.newContent && data.changes.newContent.publishReference) {
+			Object.assign(message, {
```
sjparkinson: @i-like-robots thoughts on this?

i-like-robots: This is fine by me - though it may be easier to pass the reference through explicitly rather than digging into `changes`, E.G.:

```js
logChangedContent({ uuid: item.id, ref: item.publishReference, event: 'BULK_UPDATE' })
```

This would also have the benefit of applying to new content as well.

bjfletcher:
@i-like-robots @sjparkinson I was searching on Splunk for the CREATE, UPDATE or DELETE logs, but didn't see any. I think this bulk API isn't in use?

i-like-robots:
@bjfletcher The events are `CONTENT_CREATE`, `CONTENT_UPDATE` and `CONTENT_DELETE` (these are analogous to notifications from the content API).

The bulk process is used by the background content refresh, we only log that this process has begun and completed or failed, not the items it updates.

I've added some more info the [splunk wiki page](https://github.com/Financial-Times/next-es-interface/wiki/Splunk-queries).

i-like-robots:
Do we still want this @sjparkinson or is it now covered by storing the property in ES?

sjparkinson:
Yep we still want this, so the id can be queried in Splunk. I'll work with @bjfletcher this week to fix up my test.

sjparkinson:
OK @i-like-robots this should be ready for review.

server/lib/elastic/bulk.js
```diff
@@ -23,7 +24,14 @@ module.exports = function (uuids) {
...
 
 		items.forEach((item) => {
-			item && queue.push.apply(queue, format(item));
+			if (item) {
+				queue.push.apply(queue, format(item));
+				logChangedContent({
+					uuid: item.id,
+					event: 'BULK_UPDATE',
+					publishReference: item.publishReference,
+				});
```
i-like-robots: Not sure it's worth logging this because it will create millions of entries each week that don't have much value. The publish reference shouldn't change after the initial creation anyhoo.

i-like-robots:
This looks good to me, just wondering about the value in logging each item pushed through the background update process.

PR #683, opened 2017-01-17
WIP: proposal for working with new internal API for rich journalism stuff

adgad:
@i-like-robots @kavanagh @TheoLeanse @richard-still-ft 

This PR adds a fetch to the (soon to be) newly created /internalcontent endpoint for.

This endpoint will contain all the stuff from /enrichedcontent as well as less "stable" changes that we're making for Rich Journalism.

For now, I think it makes sense to keep it completely separate and still fetch /enrichedcontent. This will allow us to make sure we're happy with both the responsiveness of the endpoint and monitor how often/frequently we decide to try/break fields, without worrying about breaking the core article content.


i-like-robots:
If these features are still in flux, might it work better for this to live in a separate app that can deliver the content JIT to the article page?

Just thinking that if the `internalcontent` API is changing faster than ES updates (once ever 7 days) then ES becomes a bottleneck.

adgad:
Potentially. Trying out a similar approach for topper - but for other things would also mean having to duplicate effectively all the stylesheets from here, and the nice stuff like resolving images. Also would slow down all article requests, so I guess it would only really be okay behind a flag.

Hoping to reduce the bottleneck by setting up a nice way of working with a local Test Methode => Test UPP => local ES interface => local ElasticSearch to article (which should reduce the amount of flux, as most of the flux should be relatively resolved locally).

And initially I suspect we'll be limiting new things to a handful of test UUIDs, so could always manually reingest.

kavanagh:
This looks good to my new eyes :-)

PR #685, opened 2017-01-17
feat: add unpublished experimental content to local ES

TheoLeanse:
@adgad @kavanagh @gvonkoss 

putting this task in ES-interface for now. 

Arjun raised the possibility of using this app's `preview` endpoint to Next-ify the content coming out of CAPI (test or mock).


TheoLeanse:
```export AWS_ACCESS_KEY_ID=[redacted];export AWS_SECRET_ACCESS_KEY=[redacted];export API_KEY=[redacted]; tasks/experimental-local-es preprod tasks/experimental-UUIDs.txt```

File: tasks/experimental-local-es [hunk 1]
```diff
@@ -0,0 +1,53 @@
+#!/usr/bin/env node
+
+require('isomorphic-fetch');
+const fetchres = require('fetchres');
+const fs = require('fs');
+
+const articleV3 = require('../../next-es-interface/server/models/article-v3');
```
adgad: This can just be `../server`?

tasks/experimental-local-es
```diff
@@ -0,0 +1,53 @@
...
+
+const elastic = 'http://localhost:9200/v3_api_v2';
```
adgad: I *think* this is now called/going to be called /content now, so probably makes sense to start using that

adgad: i.e. it's already been aliased, and will soon be renamed

PR #688, opened 2017-01-19
Fix subheading extraction

i-like-robots:
A small bugfix this one...

When the subheading falls back to selecting the first paragraph of the opening or body content it doesn't check to see if it has any text first.

@andygnewman do you know if this field is still required anyway?


andygnewman:
I don't think it is used anymore. The only practical use would be for podcasts and blogs that don't have a standfirst and the subheading was an alternative if you always wanted some extra info.

However we've disregarded all of that in teasers, so I think it can be deprecated.

i-like-robots:
Awesome

PR #689, opened 2017-01-19
Content negotiation and data sources

i-like-robots:
Currently data fetching for CAPI content is buried inside the article model. This was fine when we were dealing with a single content type but we have since hacked in videos and there are more content types on the way.

This change pulls the data fetching logic out of the models - because you cannot model without knowing what the content _is_!

The top-level function accepts a content API UUID and it will return a promise. The promise will resolve with an instance of the matching data _type_, E.G.:

```js
const dataSources = require('./server/data-sources')

// article
dataSources('3501446a-de36-11e6-86ac-f253db7791c6').then((result) => {
  console.log(result) // ArticleType
})

// video
dataSources('e0031e7d-155c-390f-a7fe-6d45d9d00f7b').then((result) => {
  console.log(result) // VideoType
})
```

The `dataSources` function itself will fetch CAPI V1 and V2 content but _types_ may be added to decorate the response with extra data. A `type` is a class with static `match` and `model` methods and a `data` instance method (this is quite like [profiles used by stream pages](https://github.com/Financial-Times/next-stream-page/wiki/profiles)).

The static `match` method will receive the CAPI responses from `dataSources` and return a boolean as to whether or not it can deal with it.

The static `model` method returns a reference to the associate model module for the type.

The `data` method should return a promise that resolves an object with properties for each data source.

```js
class MyAmazingType {
  static match ({ capiV2 }) {
    return capiV2 && capiV2.webUrl.includes('video.ft.com')
  }

  static model () {
    return require('../../models/amazing')
  }

  data () {
    // zip from `promise-patterns` module is like Promise.all() but for objects =]
    return zip(Object.extend({}, this.initialData, {
      video: fetchFromBrightcove(this.initialData.capiV2.webUrl.split('/').pop())
    }))
  }
}
```

In theory this (when combined with the [updated models](https://github.com/Financial-Times/next-es-interface/issues/553) should make it much easier going forward to add different types of content (packages, links, experimental stuff, etc.), data sources and properties.

Sanity check:

screen shot 2017-01-23 at 15 05 36


i-like-robots:
CC @adgad @TheoLeanse @andygnewman @kavanagh

server/negotiation/extensions/video.js
```diff
@@ -0,0 +1,46 @@
...
+
+const decorate = ({ capiV2 }) => {
+	const brightcoveId = capiV2.webUrl.replace('http://video.ft.com/', '');
+
+	// zip expects each value to be a function that returns a promise
+	return zip({
+		capiV2: () => Promise.resolve(capiV2),
+		brightcove: () => fetchBrightcoveData(brightcoveId),
+		videoEditor: () => fetchVideoEditorData(brightcoveId)
+	});
```
i-like-robots: https://github.com/i-like-robots/promise-patterns/pull/5 simplifies this

server/negotiation/index.js
```diff
@@ -0,0 +1,59 @@
...
+const fetchCapiJson = require('../lib/fetch-capi-json');
+const extensions = require('./extensions');
+
+const UUID = /(?:^|\/)\w{8}-\w{4}-\w{4}-\w{4}-\w{12}$/;
+
+const fetchCapiV1 = (uuid) => {
+	return fetchCapiJson(`http://api.ft.com/content/items/v1/${uuid}`)
+		.then((data) => {
+			// avoid attempting to model strange responses
+			if (data && data.item && UUID.test(data.item.id)) {
```
i-like-robots: this check can be removed as no extensions will match anyway

PR #700, opened 2017-01-25
Ingest content placeholders

TheoLeanse:
@i-like-robots @adgad


server/data-sources/types/article.js
```diff
@@ -10,9 +10,16 @@ const isFastFT = (content) => (
...
+	content &&
+	content.webUrl &&
+	!content.webUrl.match(/^https:\/\/www.ft.com/) &&
+	content.types.find(type => type === 'http://www.ft.com/ontology/content/Content')
+);
+
 class ArticleType extends Type {
 	static match ({ capiV1, capiV2 }) {
-		return isCapiV1(capiV1) || isFastFT(capiV2);
+		return isCapiV1(capiV1) || isFastFT(capiV2) || isContentPlaceholder(capiV2);
```
i-like-robots: Should this be a new content type rather than an article?

server/lib/elastic/model.js
```diff
@@ -3,16 +3,23 @@ const sanityCheck = require('../sanity-check');
...
+	let type;
 	return dataSources(uuid)
+		.then((dataType) => type = dataType)
 		.then((dataType) => {
 			return dataType.model();
 		})
 		.then((result) => {
 			// validate the content before attempting to ingest it
-			const issues = sanityCheck(result);
+			const typeValidationIssues = type.validate && type.validate(result);
```
adgad: Could probably simplify this into something like:
```
const issues = type.validate ? type.validate(result) : sanityCheck(result);
if(issues) {

}
```

server/models/placeholder/main-image-v2.js
```diff
@@ -0,0 +1,24 @@
+const logger = require('@financial-times/n-logger').default;
```
adgad: Is there a reason this file is different to the other main-image-v2 in lib?

server/lib/elastic/model.js
TheoLeanse: 👌

server/models/placeholder/main-image-v2.js
TheoLeanse: only that the alternative images which a placeholder gets has no title or description

adgad:
@TheoLeanse @i-like-robots @kavanagh I _think_ this is good to go now, if y'all approve 😁

server/models/placeholder/main-image-v2.js
```diff
@@ -0,0 +1,24 @@
...
+		ratio: image.pixelWidth && image.pixelHeight ? image.pixelWidth / image.pixelHeight : null
+	}))
+
+	.catch((err) => {
+		logger.error({
+			event: 'MAIN_IMAGE_FETCH_FAIL',
+			error: err.toString(),
+			uuid: imageId
+		});
+	})
```
adgad: I think it's slightly different to the `main-image-v2` in article - in that it's not a mainImage and is not an imageSet. But they should probably be refactored to share the same code.

adgad: I'm going to rename to alternative-image-v2 for now so it's less confusing - but images will need a bit of a redo when we come to putting the topper in.

i-like-robots:
👍

PR #702, opened 2017-01-25
Main image transforms guard for non aws sources

pixelandpage:
cc @adgad


adgad:
Looks sensible - could you add a test for the non-"com.ft.blah" case

PR #704, opened 2017-01-25
Refactor metadata extraction

i-like-robots:
This pushes the 'extraction' step back into the normalization module so that it's easier to reuse the other metadata transforms.


server/models/video.js
```diff
@@ -112,17 +112,10 @@ module.exports = ({ capiV2, brightcove, videoEditor }) => {
...
 
 		// we don't have a CAPI V1 response, only metadata
-		content.metadata = metadataV1({
-			item: {
-				metadata: {
-					sections: concordances.concat(extraTags)
-				}
-			}
-		});
+		content.metadata = metadataV1(concordances.concat(extraTags));
```
i-like-robots: 1000% easier!

PR #706, opened 2017-01-25
Fix incorrect bodyXML property check in article normalization

i-like-robots:


server/models/article/normalize-v2.js
```diff
@@ -11,7 +11,7 @@ module.exports = (article) => ({
 	standfirst: article.standfirst,
 	byline: article.byline && article.byline !== '—' ? tidyByline(article.byline) : undefined,
 	openingXML: article.openingXML ? xmlV2(article.openingXML) : undefined,
-	bodyXML: article.openingXML ? xmlV2(article.bodyXML) : undefined,
+	bodyXML: article.body ? xmlV2(article.bodyXML) : undefined,
```
adgad: Should this be bodyXML?

i-like-robots: 😴

PR #715, opened 2017-01-31
Fix placeholder metadata model

adgad:
Use the metadata-v1 model from article, and make sure that the primaryTag is set on the correct thing.

@TheoLeanse @i-like-robots


TheoLeanse:
+1

PR #716, opened 2017-01-31
Set title to alternative title if no title set

adgad:
This is to prevent myFT (and potentially other areas e.g. email, which don't use Next API/n-teaser and therefore still need a title field) from breaking.

@TheoLeanse @i-like-robots


TheoLeanse:
+1

PR #719, opened 2017-01-31
Add a type for all the models

adgad:
@i-like-robots @TheoLeanse @andygnewman @kavanagh


i-like-robots:
👍

i-like-robots:
CC @ironsidevsquincy + @andygout does this mean changes are required for Next API?

TheoLeanse:
🕺

ironsidevsquincy:
Yes! Means we can stop doing https://github.com/Financial-Times/next-api/blob/master/server/lib/backends/capi/utils/resolve-content-type.js

ironsidevsquincy:
Well, -ish. Will there be a `liveBlog` type?

adgad:
I think the `type` should be matching the model in ES, so probably will have to leave that `liveBlog` check in Next API (or maybe have an isLiveBlog field or something along those lines in ES?)

i-like-robots:
I reckon leave it to next API as it's just some sugar over `realtime: true` (as per CAPI)

PR #721, opened 2017-01-31
Make way for HTML transcriptions

commuterjoy:
Moving some mappings around,

```
brightcove.description -> standfirst
brightcove.long_description -> alternativeStandfirsts.promotionalStandfirst
videoEditor.transcription.html -> bodyHTML
```


server/models/video.js
```diff
@@ -26,11 +27,11 @@ function video ({ capiV2, brightcove, videoEditor }) {
 
 	model.set('alternativeTitles', normalV2.alternativeTitles || {});
 
-	model.set('standfirst', brightcove.metadata.description);
+	model.set('standfirst', brightcove.metadata.long_description);
 
 	model.set('byline', normalV2.byline);
 
-	model.set('bodyHTML', `<p>${brightcove.metadata.long_description}</p>`);
+	model.set('bodyHTML', extractTranscription(videoEditor));
```
i-like-robots: This will need a fallback as content is rejected without a body.

i-like-robots:
Does the base branch for this PR need changing?

commuterjoy:
@i-like-robots ok, i think this is right

server/models/video.js
```diff
@@ -28,9 +29,13 @@ function video ({ capiV2, brightcove, videoEditor }) {
 
 	model.set('standfirst', brightcove.metadata.description);
 
+	model.set('alternativeStandfirsts', {
+		promotionalStandfirst: brightcove.metadata.long_description || {}
```
i-like-robots: `... || undefined` might work better because this gets turned into a JSON string

commuterjoy: ok boss. done.

PR #723, opened 2017-02-02
Get HTTPS video sources

ironsidevsquincy:
Because they're not duplicates, but HTTP/S. Durrrrrrr


i-like-robots:
OK! Why would non secure even by an option? If they're the same file why... Oh, whatever...

🚢

PR #727, opened 2017-02-06
Internal content take two

adgad:
@TheoLeanse @i-like-robots @gvonkoss @kavanagh 

Adding the topper to elastic search will help resolve the performance issues currently on the rich pages (resolving the /internalcontent and then all the images slows down article responses), and will make spoor enrichment easier.

* Redo hitting _what will be_ the internalcontent endpoint as an additional annotation for articles
* Add topper to the response (not adding to the schema yet as it's still in flux)
* Refactor v2 image code so articles/toppers/placeholders all use the same thing

Do not merge until the api exists and it is verified working!


TheoLeanse:
Looks good

adgad:
Blah, I was hoping we could do this without having to reindex but just realised we can't (because images is a nested field). Given this is still likely to change, I think perhaps we should hold back and keep doing it the article way (and accept it will be a tad slower)

schema/content.json
```diff
@@ -274,6 +274,10 @@
 				"realtime":{
 					"type":"boolean"
 				},
+				"topper": {
```
adgad: @i-like-robots @TheoLeanse I know this is a bit cheeky, but it seems to do the trick for now.

Tested adding a topper to the existing schema, and created a new schema with this empty topper and it creates successfully and stores the topper in the right format.

server/models/article/image-v2.js
```diff
@@ -2,25 +2,22 @@ const logger = require('@financial-times/n-logger').default;
...
-				error: err.toString(),
-				uuid: imageId
-			});
-		})
+				uuid: image.id
+			}, err);
+		});
 	}
-
 };
```
i-like-robots: Probably worth a TODO to merge this with https://github.com/Financial-Times/next-es-interface/blob/master/server/models/article/main-image-v2.js

server/models/article/imageset-v2.js
```diff
@@ -0,0 +1,22 @@
...
+		))
+		.then(imageV2)
+		.catch((err) => {
+			logger.error({
+				event: 'MAIN_IMAGE_FETCH_FAIL',
+				error: err.toString(),
+				uuid: imageset.id
+			});
+		});
+	}
```
i-like-robots: As above

i-like-robots: NVM, I see what's happening now...

schema/content.json
i-like-robots: it won't do any harm =]

PR #731, opened 2017-02-10
Add brightcove URL to provenance field if article has embedded video

i-like-robots:
Reviving this old feature temporarily until we come up with a better way of finding content with videos embedded in it. We need it to report some stats.

Currently it would involve a query like so (because we are [tokenising](https://www.elastic.co/guide/en/elasticsearch/reference/1.5/analysis-standard-tokenizer.html) the body content -- uselessly as we don't actually search... rant rant -- and a query may only target individual tokens):

```json
{
    "query": {
        "span_near": {
        	"clauses": [
        		{ "span_term": { "bodyHTML": "n" } },
        		{ "span_term": { "bodyHTML": "content" } },
        		{ "span_term": { "bodyHTML": "video" } },
        		{ "span_term": { "bodyHTML": "brightcove" } }
        	],
        	"slop" : 0,
        	"in_order" : true
        }
    }
}
```

So this hack will re-add the Brightcove API URL to the provenance field if there's a video embedded in the article.


i-like-robots:
@adgad @kavanagh would it be useful to have a CAPI V1 style `assets` field that says if content contains images and/or videos.

PR #736, opened 2017-02-14
Respect isPrimarilyClassifiedBy for placeholder tags

adgad:
@TheoLeanse @i-like-robots @i-like-robots 

This probably needs a bit of a rethink at some point to more closely match `display-tags.js` logic for articles, but it should fix the immediate problem.


TheoLeanse:
+1

i-like-robots:
Any docs for this stuff?

Issue #733, opened 2017-02-13
Stop analysing HTML fields


i-like-robots:
We current let the default analyser go to work on the `openingHTML` and `bodyHTML` fields and we shouldn't, because content like this:

```html
<div class="n-content-video n-content-video--brightcove"><a href="http://video.ft.com/4714974563001" data-embedded="true" data-asset-type="video"></a></div>
```

Gets tokenised into the following:

```json
{
  "tokens": [
    {
      "token": "div",
      ...
    },
    {
      "token": "class",
      ...
    },
    {
      "token": "n",
      ...
    },
    {
      "token": "content",
      ...
    },
    {
      "token": "video",
      ...
    },
    {
      "token": "n",
      ...
    },
    {
      "token": "content",
      ...
    },
    {
      "token": "video",
      ...
    },
    {
      "token": "brightcove",
      ...
    },
    {
      "token": "a",
      ...
    },
    {
      "token": "href",
      ...
    },
    {
      "token": "http",
      ...
    },
    {
      "token": "video.ft.com",
      ...
    },
    {
      "token": "4714974563001",
      ...
    },
    {
      "token": "data",
      ...
    },
    {
      "token": "embedded",
      ...
    },
    {
      "token": "true",
      ...
    },
    {
      "token": "data",
      ...
    },
    {
      "token": "asset",
      ...
    },
    {
      "token": "type",
      ...
    },
    {
      "token": "video",
      ...
    },
    {
      "token": "a",
      ...
    },
    {
      "token": "div",
      ...
    }
  ]
}
```

Which is very cumbersome (though [not impossible](https://www.elastic.co/guide/en/elasticsearch/reference/1.5/query-dsl-span-near-query.html)) to query.

We should update the schema to set our HTML fields to be non-analysed so we may query them using simple wildcard strings when we need to.

PR #738, opened 2017-02-14
Content stats field

i-like-robots:
Unfortunately #731 did not scratch the itch.

This PR adds simple ~boolean flags~ feature counts ([inspired by Spoor](https://github.com/Financial-Times/spoor-enrichment/blob/a6a6a7f30976a5fc530441151c993678cacc114a/server/transforms/content-api-v1.js#L72)) when ingesting content so we may easily query what articles _contain_. E.G.

~http://next-elastic.ft.com/content/item/_search?q=type:article+AND+contentFlags.hasVideos:true~
http://next-elastic.ft.com/content/item/_search?q=type:article+AND+contentStats.videos:>0

Somewhat appeases #733.


schema/content.json
```diff
@@ -263,6 +263,22 @@
 						}
 					}
 				},
+				"contentFlags":{
+					"properties":{
+						"hasRelatedBoxes":{
```
adgad: Would be nice if these were integers actually, so we can see how many an article has

i-like-robots: I don't mind that, who else would have an interest... @kavanagh @andygnewman... booleans or counts?

i-like-robots: I'll rename this stats if we go with numbers

PR #744, opened 2017-02-15
Fix ML publishing issues: omit check if bodyHTML exists.

### LINKED ISSUE #742: Allow realtime articles to pass validation without body content
Live blogs may be published without body content from CAPI V2. We believe we worked around this previously by throwing out content without a V1 response but now that Alphaville and FastFT blog posts skip this requirement they're falling foul of this validation rule.

roland-vachter:
Fixes #742


server/data-sources/types/article.js
```diff
@@ -20,6 +21,10 @@ const isAlphaville = (content) => (
 	content && content.types && content.types.find((type) => type === ARTICLE_TYPE) && content.webUrl && content.webUrl.includes('ftalphaville.ft.com')
 );
 
+const isMarketsLive = (content) => (
+	isAlphaville(content) && content.webUrl.includes('/marketslive/')
+);
```
i-like-robots: Should we just base this decision on the `realtime` field? My understanding is that all live blogs may suffer from this issue

roland-vachter: I see "realtime": false for MarketsLive, but maybe that's just for closed sessions? Maybe it's true while it's on?

davidlintonattheft: Where do you see this?
this marketslive is realtime: true https://ftalphaville.ft.com/marketslive/2017-02-15/
i.e. 66f2a2cf-987b-3bcf-b58c-6eb7f5ccab0e

realtime doesn't change when it is closed it should always be true for a live blog

davidlintonattheft: At the very least you should allow realtime:true through. I guess I don't care if you add an extra hack specially for markets live.

roland-vachter: I could find a couple of examples from 2015, also the test published today, but I think it's related to the comments.enabled field as well. When that is false, realtime is false, but for articles just published it should be true.

davidlintonattheft: ok that's a bug which we should fix.

roland-vachter:
Updated the rule to check all realtime content.

i-like-robots:
Thanks, let's 🚢 it

PR #748, opened 2017-02-16
[WIP] modelling for series/packages

gvonkoss:
@adgad @kavanagh @TheoLeanse @i-like-robots


test/server/data-sources/types/article-spec.js
```diff
@@ -69,15 +69,21 @@ describe('Data source - article', () => {
...
+				.get(`/best-books/${FIXTURE_V2_ID}`)
+				.times(1)
+				.reply(500, null);
+
 			instance = new Subject({ capiV1: fixtureV1, capiV2: fixtureV2 });
 
 			return instance.data({ capiV1: fixtureV1, capiV2: fixtureV2 }).then((result) => {
-				expect(nock.isDone()).to.be.true;
+				expect(internal.isDone()).to.be.true;
+				expect(mock.isDone()).to.be.true;
```
i-like-robots: There's not an advantage in this case of assigning each mock and calling `isDone` _but_ if you do this:

```diff
--expect(internal.isDone()).to.be.true;
--expect(mock.isDone()).to.be.true;
++internal.done()
--mock.done()
```

You will get a detailed failure rather than just a boolean.

schema/content.json
```diff
@@ -303,6 +303,10 @@
 					"properties": {
 					}
 				},
+				"descriptionHTML": {
```
adgad: Have you tried creating an index with this schema? I suspect it will break...

server/data-sources/index.js
```diff
@@ -16,6 +18,35 @@ const fetchCapiV2 = (uuid) => (
...
+
+//TODO remove this when internalcontent has data
+// fetches mock data to add to internal content
+const fetchCapiInternal = (uuid) => {
+	return Promise.all([internal(uuid), mock(uuid)])
+	.then((response) => {
+		return Object.assign(response[0], response[1])
+	})
+	.catch((error) => {
+		logger.warn({ event: 'MULTIPLE_FETCH_FAIL', uuid, status: error.status, requestId: error.requestId, error: error.error });
```
adgad: Maybe make that event a bit less scary! 😬

server/data-sources/types/article.js
```diff
@@ -48,16 +36,15 @@ class ArticleType extends Type {
 	static check (content) {
 		if (content.realtime === true) {
 			// CAPI V2 will omit content body for realtime content.
-			const tests = Object.keys(validateFields.tests).filter((item) => item !== 'bodyHTML');
-			return validateFields(content, tests);
+			return validateFields(content, ['id', 'publishedDate', 'metadata']);
 		} else {
-			return validateFields(content);
+			return validateFields(content, ['id', 'publishedDate', 'metadata', 'bodyHTML']);
```
adgad: What's the reason for this change?

adgad: Oh never mind, it's because we don't want to validate the 'contains' field

server/lib/fetch-mock-json.js
```diff
@@ -0,0 +1,23 @@
...
+	return fetch(endpoint)
+		.then((response) => {
+			if (response.ok) {
+				return response.json();
+			}
+			return response.text()
+				.then((text) => {
+					const requestId = response.headers.get('x-request-id') || 'UNKNOWN';
+					const server = response.headers.get('server') || 'UNKNOWN';
+					const masheryMessageId = response.headers.get('x-mashery-message-id') || 'UNKNOWN';
```
adgad: don't think we need these headers for the S3 fetch

server/models/article.js
```diff
@@ -148,6 +148,18 @@ function article ({ capiV1, capiV2, capiInternal }) {
 		model.set('topper', topper(capiInternal));
 	}
 
+	if (capiInternal && capiInternal.containedIn) {
+		model.set('containedIn', capiInternal.containedIn);
+	}
+
+	//TODO: add a type for packages; apply conditional with what comes out of capiInternal
```
adgad: Can we delete this block if it's not relevant anymore.

server/models/article/rich-journalism/content-package.js
```diff
@@ -0,0 +1,13 @@
+module.exports = (article) => {
```
adgad: It doesn't look like this file is being used anywhere?

server/data-sources/index.js
```diff
@@ -16,6 +18,35 @@ const fetchCapiV2 = (uuid) => (
...
+	// resolves promise if no mock available for uuid
+	.catch(() => { return null })
+);
+
+//TODO remove this when internalcontent has data
+// fetches mock data to add to internal content
+const fetchCapiInternal = (uuid) => {
+	return Promise.all([internal(uuid), mock(uuid)])
+	.then((response) => {
+		if (response[1]) logger.info({ event: 'MOCK_FETCH_SUCCESSFUL' });
```
adgad: should probably add the UUID to this event

PR #749, opened 2017-02-16
Log conetnt type when ES content is created or updated

ironsidevsquincy:


server/lib/elastic/update.js
```diff
@@ -49,6 +50,8 @@ module.exports = function (uuid) {
 		get(uuid)
 	])
 		.then(([ newContent, oldContent ]) => {
+			const event = oldContent ? 'ES_CONTENT_UPDATED' : 'ES_CONTENT_CREATED';
+			logger.info({event, uuid, content_type: newContent.type });
```
i-like-robots: The reindex and create methods will both log already (`CONTENT_CREATE` and `CONTENT_UPDATE`) - do we just need to add the type field?

i-like-robots: For reference, that's here: https://github.com/Financial-Times/next-es-interface/blob/master/server/lib/log-changed-content.js#L20

PR #755, opened 2017-02-21
Fix placeholder type missing last updated time field

i-like-robots:
This means that this content type is not subject to the continuous updater =[

@adgad @gvonkoss


adgad:
👍

i-like-robots:
We need to reingest a few things, not too many http://next-elastic.ft.com/content/item/_search?q=_missing_:_lastUpdatedDateTime

PR #756, opened 2017-02-22
Fix MarketsLive ingesting when it's missing from CAPI

roland-vachter:
MarketsLive need the tag 'Markets' together with 'FT Alphaville', when it is ingested with the hack (missing from CAPI v1), they should have the Markets tag added as well.


server/models/article.js
```diff
@@ -101,6 +101,12 @@ function article ({ capiV1, capiV2, capiInternal }) {
 		model.get('metadata').then((tags) => {
 			model.set('metadata', alphaville(merged.id, tags));
 		});
+
+		if (merged.webUrl.includes('marketslive')) {
+			model.get('metadata').then((tags) => {
+				model.set('metadata', marketslive(merged.id, tags));
+			});
```
i-like-robots: Could this be rolled into the alphaville method as it's the same content source?

roland-vachter: Ok

i-like-robots:
LGTM! Maybe squash first =]

PR #760, opened 2017-02-28
Use correct primary theme and section for placeholders

adgad:
@TheoLeanse @i-like-robots @gvonkoss 

This moves the whole "picking primary theme/section" for placeholders logic into the concording - which means it should go through the same article metadata logic and pick up the correct display tags.


TheoLeanse:
LGTM

PR #770, opened 2017-03-02
Remove metadata validation from video content

ironsidevsquincy:
As we get tags from Brightcove, via the mapping spreadsheet, it's possible that the content ends up with no valid TME metadata


i-like-robots:
So dirty feeling!

PR #773, opened 2017-03-03
Don't duplicate feature.blogposts query string

adgad:
@i-like-robots @TheoLeanse @kavanagh


i-like-robots:
LGTM, thank you again for doing it. I might come back through here later and parse the URL etc for extra safety but let's :shipit:!

PR #763, opened 2017-03-01
Elasticsearch 5.1

sjparkinson:
🚨  Don't merge this yet.

Merge of https://github.com/Financial-Times/next-elasticsearch back into this repository.

To Do:

- [ ] Get the new clusters in sync
- [ ] Plan and switch to the new clusters
- [ ] Clean up all the next-elasticsearch stuff

PR #779, opened 2017-03-06
Topper update

gvonkoss:
@adgad does this look right?


server/models/article/rich-journalism/image-set.js
```diff
@@ -0,0 +1,19 @@
+const imageV2 = require('../image-v2');
+
+module.exports = (article) => {
+	//TODO: remove article.topper when prod complete
+	const imageSet = article.topper.images || article.leadImages;
```
adgad: imageSet is a bad name here unfortunately, because imagesets are a CAPI thing. Maybe just call it lead-images?

adgad: And then rather than checking `const imageSet = article.topper.images || article.leadImages;` in here, you can pass the right one in as the function parameter

server/models/article/rich-journalism/topper.js
```diff
@@ -1,28 +1,24 @@
...
+//TODO remove article.topper.theme when prod complete (topper.bgColor?)
 module.exports = (article) => {
 	if(article && article.topper && article.topper.theme !== '') {
-
 		const model = {
 			headline: article.topper.headline || article.title,
 			standfirst: article.topper.standfirst || article.standfirst,
-			theme: article.topper.theme,
-			backgroundColour: article.topper.backgroundColour || article.topper.bgColor
+			layout: article.topper.theme || article.topper.layout,
```
adgad: I think we need to have both theme and layout as separate fields here, otherwise we'll break existing toppers.

server/models/article/rich-journalism/topper.js
```diff
@@ -1,28 +1,24 @@
...
+			backgroundColour: article.topper.backgroundColour || article.topper.bgColor,
```
adgad: Trailing comma

server/models/article/rich-journalism/topper.js
```diff
@@ -1,28 +1,24 @@
...
-				article.topper.images.map(imageRef => {
-					return imageV2(imageRef)
-						.then(img => Object.assign({}, img, { type: imageRef.type }));
-				}))
-				.then(images => {
-					model.images = images;
-					return model;
-				});
+//TODO: delete images when prod is ready
+		if(article.topper.images || article.leadImages) {
```
adgad: We don't really need to be using article.leadImages here, because we can do that mapping downstream in the article (which will make it easier to just delete this code when we're done.

test/server/models/article/rich-journalism/topper-spec.js
```diff
@@ -1,28 +1,25 @@
...
 describe('Topper', () => {
 	it('Sets the headline, background colour, theme', () => {
-		return subject(fixture).then(result => {
-
+		subject(fixture).then(result => {
 			expect(result.headline).to.equal('Topper headline');
 			expect(result.standfirst).to.equal('Topper standfirst');
 			expect(result.backgroundColour).to.equal('black');
 			expect(result.images.length).to.equal(3);
-			expect(result.images[0].id).to.equal(123);
```
adgad: These should still be present on the topper, no?

schema/content.json
```diff
@@ -299,6 +299,9 @@
 				"realtime":{
 					"type":"boolean"
 				},
+				"design":{
+					"properties": {}
+				},
 				"topper": {
```
adgad: Don't we need leadImages here as well? I assume it wouldn't work without it...

schema/content.json
```diff
@@ -307,6 +310,10 @@
 					"type":"string",
 					"index":"not_analyzed"
 				},
+				"leadImages": [],
+				"tableOfContents": {
+					"properties": {}
+				},
```
i-like-robots: Just a heads up we've changed the schema quite a lot in #763 which will bulldoze this - let's get this set up so we don't lose it

gvonkoss: oh no... am I going to have to rebase again!?

i-like-robots: Depends who gets there first, I imagine #763 will ship this week

gvonkoss: I'm _so_ beating you to it

i-like-robots: Let's get this added to the new schema in full, are `design`, `topper` and `tableOfContents` simply strings? Is `leadImages` just an array of strings?

sjparkinson: And the winner is... not me 😢

PR #803, opened 2017-03-23
Video v2 main image

ironsidevsquincy:
Will now come through as `MediaResource` content, so update the transform accordingly

Also added tests for the other video v2 helpers attachments and storyPackage


server/models/video-v2.js
```diff
@@ -15,7 +15,7 @@ const resolveMetadata = (annotations = [], dependencies = {}) => {
 };
 
 module.exports = ({ capiV2 = {} } = {}, dependencies = {}) => {
-	const deps = Object.assign({ concord, imagesetV2, storyPackage }, dependencies);
+	const deps = Object.assign({ concord, mainImage, storyPackage }, dependencies);
```
i-like-robots: Just a heads up that this app tends to use proxyquire rather than DI

server/models/video-v2.js
```diff
@@ -2,8 +2,8 @@ const Model = require('./model');
 const metadataV1 = require('./article/metadata-v1');
 const annotations = require('./article/annotations');
 const normalizeV2 = require('./article/normalize-v2');
-const imagesetV2 = require('./article/imageset-v2');
 const attachments = require('./video/attachments');
+const mainImage = require('./video/main-image');
```
i-like-robots: looks like this functionality already exists (just a bit unexpectedly named) with only slight differences. Maybe worth checking out the two can be reconciled?

https://github.com/Financial-Times/next-es-interface/blob/master/server/models/article/image-v2.js

server/models/video-v2.js
ironsidevsquincy: proxyquire is way too much like magic, but can switch over if you prefer

server/models/video-v2.js
ironsidevsquincy: Ah, will use that instead

server/models/video-v2.js
i-like-robots: I'm not really fussy whatever works for you

PR #804, opened 2017-03-27
Check all matched annotations for primary tag classification

adgad:
@ironsidevsquincy @i-like-robots @bjfletcher


i-like-robots:
Ah, my bad! Broken in #801

i-like-robots:
Have raised this with the UPP team and as I think this is not a very useful feature.

i-like-robots:
It's a feature, so we'll have to do some work in future to handle it - though it could get quite complicated.

https://sites.google.com/a/ft.com/universal-publishing/documentation/introduction-to-metadata/annotations/triple-annotations?pli=1

